PR #7, opened 2017-09-08
Field reordering

madtrick:
Extend the migration language with methods to move a field within a CT. Moving a field within a CT has impact in how the field of an entry of the edited content type are later displayed in Contentful Webapp (i.e. fields moved to top will be displayed on top when editing an entry and so on).

If fields are not reordered they will be displayed in the order they are added.

Plan output:

```bash
The following migration has been planned

Update Content Type food

  Create field calories
    - type: "Number"
    - name: "How many calories does it have?"
  Move field calories to the first position

  Create field sugar
    - type: "Number"
    - name: "Amount of sugar"
  Move field sugar to the last position

  Create field vegan
    - type: "Boolean"
    - name: "Vegan friendly"

  Create field producer
    - type: "Symbol"
    - name: "Food producer"
  Move field producer before field vegan

  Create field gmo
    - type: "Boolean"
    - name: "Genetically modified food"
  Move field gmo after field vegan
```

### DIFF
--- a/lib/migration-chunks/validation/errors.js
+++ b/lib/migration-chunks/validation/errors.js
@@ -31,6 +31,26 @@ module.exports = {
       CONTENT_TYPE_DOES_NOT_EXIST: (id, ctId) => {
         return `You cannot set a property on field with id "${id}" on content type "${ctId}" because it does not exist.`;
       }
+    },
+    move: {
+      FIELD_DELETED: (id) => {
+        return `You cannot move the field with id "${id}" because it has already been deleted.`;
+      },
+      FIELD_DOES_NOT_EXIST: (id) => {
+        return `You cannot move the field with id "${id}" because it does not exist.`;
+      },
+      FIELD_ALREADY_MOVED: (id) => {
+        return `You cannot move the field with id "${id}" more than once per migration.`;
+      },
+      CONTENT_TYPE_DOES_NOT_EXIST: (id, ctId) => {
+        return `You cannot move a field on content type "${ctId}" because it does not exist.`;
+      },
+      PIVOT_FIELD_DOES_NOT_EXIST: (id, direction) => {
+        return `You cannot move a field ${direction} the field "${id}" because it doesn't exist`;
+      },
+      PIVOT_FIELD_DELETED: (id, direction) => {
+        return `You cannot move a field ${direction} the field "${id}" because it has been already deleted`;
+      }
     }
   },
   contentType: {
--- a/lib/migration-chunks/validation/field.js
+++ b/lib/migration-chunks/validation/field.js
@@ -10,6 +10,152 @@ const invalidActionError = (message, step) => {
   };
 };
 
+const RELATIVE_MOVEMENTS = ['afterField', 'beforeField'];
+
+function fieldId (step) {
+  return step.payload.fieldId;
+}
+
+function pivotId (step) {
+  return step.payload.movement.pivot;
+}
+
+function contentTypeId (step) {
+  return step.payload.contentTypeId;
+}
+
+function fieldMovementKey (step) {
+  return [contentTypeId(step), fieldId(step)].join('/');
+}
+
+function fieldExists (validationContext, fieldId) {
+  return validationContext.fieldSet.has(fieldId);
+}
+
+function fieldHasBeenRemoved (validationContext, fieldId) {
+  return validationContext.fieldRemovals.has(fieldId);
+}
+
+function fieldHasBeenMoved (validationContext, movementKey) {
+  return validationContext.fieldMovements.has(movementKey);
+}
+
+const checks = {
+  'field/create': (errors, step, validationContext) => {
+    const id = fieldId(step);
+    const exists = fieldExists(validationContext, id);
+
+    if (!exists) {
+      validationContext.fieldSet.add(id);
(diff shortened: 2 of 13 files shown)


TimBeyer:
How would we display this in the plan overview?

lib/migration-chunks/validation/field.js
```diff
@@ -10,6 +10,122 @@ const invalidActionError = (message, step) => {
   };
 };
 
+function fieldId (step) {
```
TimBeyer: My personal preference would be to enforce `func-style: expression` in the repo to keep things consistent across the board since we already enforce `no-use-before-define` and function declarations don't really serve any other purpose

madtrick:
> How would we display this in the plan overview?

@TimBeyer field movement is now reflected on the plan PTAL

lib/migration-steps/validation.js
```diff
@@ -81,6 +80,64 @@ const getFieldErrors = getStepErrors.bind(null, {
...
 
+const fieldMovementValidations = {
+  toTheTop: Joi.any(),
+  toTheBottom: Joi.any(),
+  afterField: Joi.string().required(),
+  beforeField: Joi.string().required()
+};
+
+const getFieldMovementErrors = function (step) {
+  const validMoves = ['toTheTop', 'toTheBottom', 'afterField', 'beforeField'];
```
zcei: Shall we derive this from the `Object.keys` of `fieldMovementValidations`?

lib/migration-chunks/validation/field.js
```diff
@@ -26,23 +142,14 @@ module.exports = function (chunks, contentTypes = []) {
     const fieldSet = contentTypeFields[contentTypeId] || new Set();
     const fieldRemovals = recentlyRemoved[contentTypeId] || new Set();
 
+    // TODO catch movements across chunks
+    const validationContext = { fieldSet, fieldRemovals, fieldMovements: new Set() };
```
zcei: Should we really track this across chunks? And if so, are there implications that restrain us from using the same behavior as for the `contentTypeFields` / `recentlyRemoved`?

```js
const fieldMoves = recentlyMoved[contentTypeId] || new Set();
// ...
recentlyMoved[contentTypeId] = fieldMoves;
```

lib/migration-steps/validation.js
madtrick: Sounds good

madtrick: changed

lib/migration-chunks/validation/errors.js
```diff
@@ -31,6 +31,20 @@ module.exports = {
...
+      FIELD_DOES_NOT_EXIST: (id) => {
+        return `You cannot move the field with id "${id}" because it does not exist.`;
+      },
+      FIELD_ALREADY_MOVED: (id) => {
+        return `You cannot move the field with id "${id}" more than once per migration.`;
+      },
+      CONTENT_TYPE_DOES_NOT_EXIST: (id, ctId) => {
+        return `You cannot move a field on content type "${ctId}" because it does not exist.`;
+      }
     }
```
jfr3000: do we check if users are trying to move a field before or after a field which does not exist?

madtrick: :10x: catch. Fixed.

lib/migration-chunks/validation/field.js
madtrick: I think we should restrict the field movements to only once per field per migration. I can't really think of a case where it would bring any value to move the same field multiple times.

I updated the code to reflect your proposal : )

lib/migration-chunks/validation/field.js
```diff
@@ -10,6 +10,152 @@ const invalidActionError = (message, step) => {
...
+function fieldHasBeenRemoved (validationContext, fieldId) {
+  return validationContext.fieldRemovals.has(fieldId);
+}
+
+function fieldHasBeenMoved (validationContext, movementKey) {
+  return validationContext.fieldMovements.has(movementKey);
+}
+
+const checks = {
+  'field/create' (errors, step, validationContext) {
```
zcei: What was the decision now on this func-style? I find it quite confusing, but could live with it - we're not using this style anywhere else, though 🤔

lib/migration-steps/errors.js
```diff
@@ -9,6 +9,18 @@ const errors = {
...
     return `"${actualType}" is not a valid type for the ${typeName} property "${propName}". Expected "${expectedType}".`;
+  },
+  INVALID_MOVEMENT_TYPE: (typeName) => {
+    return `"${typeName}" is not a valid type for field movement. Expected "string".`;
+  },
+  INVALID_MOVEMENT_WITH_SELF: (fieldId) => {
+    return `You cannot move the field "${fieldId}" relative to itself.`;
+  },
+  INVALID_MOVEMENT_NAME: (movement) => {
+    return `"${movement}" is not a valid field movement.`
```
zcei: Missing semicolon, Travis complains 🙃

lib/migration-chunks/validation/field.js
madtrick: ah, that was me being fancy. Actually I don't know why I used it when I don't particularly like it. Will change it : )

madtrick:
@zcei fixec, PTAL

PR #10, opened 2017-09-18
Change field id

jfr3000:
This adds a `.changeFieldId()` method to content types, allowing users to change a field's ID.

It validates and covers the cases mentioned in `validation.md`.

### DIFF
--- a/lib/migration-chunks/validation/field.js
+++ b/lib/migration-chunks/validation/field.js
@@ -40,6 +40,10 @@ function fieldHasBeenMoved (validationContext, movementKey) {
   return validationContext.fieldMovements.has(movementKey);
 }
 
+function idHasBeenChangedBefore (validationContext, oldId) {
+  return Array.from(validationContext.fieldIdChanges.values()).includes(oldId);
+}
+
 const checks = {
   'field/create': (errors, step, validationContext) => {
     const id = fieldId(step);
@@ -56,12 +60,18 @@ const checks = {
         step
       ));
     }
+
+    if (validationContext.fieldIdChanges.has(id)) {
+      // a new field is created with the old id, so it's ok to use again
+      validationContext.fieldIdChanges.delete(id);
+    }
   },
 
   'field/update': (errors, step, validationContext) => {
     const id = fieldId(step);
     const exists = fieldExists(validationContext, id);
     const removed = fieldHasBeenRemoved(validationContext, id);
+    const changedId = validationContext.fieldIdChanges.get(id);
 
     if (!exists && removed) {
       errors.push(invalidActionError(
@@ -70,7 +80,12 @@ const checks = {
       ));
     }
 
-    if (!exists && !removed) {
+    if (changedId) {
+      errors.push(invalidActionError(
+        fieldErrors.changeId.REFERENCE_TO_OLD_ID(id, changedId),
+        step
+      ));
+    } else if (!exists && !removed) {
       errors.push(invalidActionError(
         fieldErrors.update.FIELD_DOES_NOT_EXIST(id),
         step
@@ -153,6 +168,39 @@ const checks = {
         step
       ));
     }
+  },
+
+  'field/rename': (errors, step, validationContext) => {
+    const oldId = step.payload.fieldId;
+    const newId = step.payload.props.newId;
+
+    if (idHasBeenChangedBefore(validationContext, oldId)) {
+      errors.push(invalidActionError(
+        fieldErrors.changeId.NO_MULTIPLE_ID_CHANGES(oldId),
+        step
+      ));
+    }
+
+    if (oldId === newId) {
+      errors.push(invalidActionError(
+        fieldErrors.changeId.ID_MUST_BE_DIFFERENT(oldId),
+        step
+      ));
+    } else if (fieldExists(validationContext, newId)) {
+      errors.push(invalidActionError(
+        fieldErrors.changeId.ID_ALREADY_EXISTS(
+          oldId,
+          newId,
+          step.payload.contentTypeId
+        ),
+        step
+      ));
+    }
+    // remember this field's old id, so that we can check if it is erroneously used later on
+    validationContext.fieldIdChanges.set(oldId, newId);
+
+    validationContext.fieldSet.delete(oldId);
+    validationContext.fieldSet.add(newId);
   }
 };
 
@@ -166,6 +214,7 @@ module.exports = function (chunks, contentTypes = []) {
 
   const recentlyRemoved = {};
   const recentlyMoved = {};
+  const changedFieldIds = {};
 
   for (const chunk of chunks) {
     const contentTypeId = chunk[0].payload.contentTypeId;
@@ -173,10 +222,9 @@ module.exports = function (chunks, contentTypes = []) {
     const fieldSet = contentTypeFields[contentTypeId] || new Set();
     const fieldRemovals = recentlyRemoved[contentTypeId] || new Set();
     const fieldMovements = recentlyMoved[contentTypeId] || new Set();
+    const fieldIdChanges = changedFieldIds[contentTypeId] || new Map();
+    const validationContext = { fieldSet, fieldRemovals, fieldMovements, fieldIdChanges };
 
-    const validationContext = { fieldSet, fieldRemovals, fieldMovements };
-
-    // ...
     const fieldSteps = chunk.filter((step) => step.type.startsWith('field'));
 
     for (const step of fieldSteps) {
@@ -197,6 +245,7 @@ module.exports = function (chunks, contentTypes = []) {
     contentTypeFields[contentTypeId] = fieldSet;
     recentlyRemoved[contentTypeId] = fieldRemovals;
     recentlyMoved[contentTypeId] = fieldMovements;
+    changedFieldIds[contentTypeId] = fieldIdChanges;
   }
 
   return errors;
--- a/lib/migration-payloads/index.js
+++ b/lib/migration-payloads/index.js
@@ -21,7 +21,7 @@ function buildFields (result, actions) {
       ));
     }
 
-    if (action.type === 'field/update') {
+    if (action.type === 'field/update' || action.type === 'field/rename') {
       const source = _.find(result.payload.fields, { id });
 
       Object.assign(source, props);
@@ -76,6 +76,21 @@ function builder (chunk, contentType) {
   return result;
 };
 
+function updateFieldsIdsForChanges (payload, fieldIdChanges) {
+  const updatedCt = _.cloneDeep(payload);
+  const fields = updatedCt.payload.fields;
+  if (fields.some((field) => field.newId)) {
+    fieldIdChanges[payload.meta.contentTypeId] = _.cloneDeep(payload);
+  }
+  for (const field of fields) {
+    if (field.newId) {
+      field.id = field.newId;
+      delete field.newId;
+    }
+  }
+  return updatedCt;
+}
+
 module.exports = function payloadBuilder (plan, contentTypes = []) {
   const ctPayloads = contentTypes.map((ct) => {
     return {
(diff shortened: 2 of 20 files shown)


lib/migration-payloads/validation/schema-validation.js
```diff
@@ -55,8 +55,12 @@ const itemsValid = Joi.object().keys({
 
 const fieldSchema = Joi.object().keys({
   id: Joi.string().required(),
+  newId: Joi.string()
+    .invalid(Joi.ref('id'))
+    .max(64)
+    // the empty case will be caught by joi by default, we don't want duplicate errors
+    .regex(/^$|^[a-zA-Z][a-zA-Z0-9_]*$/),
```
jfr3000: I tried using joi's `string.allow('')` method, but then neither `regex(..)` nor `min(1)` get applied and an empty string is considered valid...

lib/migration-payloads/validation/index.js
```diff
@@ -33,6 +33,23 @@ const checkIfDisplayFieldsExist = function (payload) {
   return errors;
 };
 
+const checkIfNewIdIsUnique = function (payload) {
+  const errors = [];
+  const fields = payload.payload.fields;
+  const fieldsWithNewIds = fields.filter((field) => field.newId);
+  fieldsWithNewIds.forEach((field) => {
```
zcei: Can we have a `for-of` or `.reduce()` here? List functionals with side-effects scare me 👻

lib/migration-steps/validation.js
```diff
@@ -14,6 +14,7 @@ const contentTypePropsValidations = {
 
 const fieldPropsValidations = {
   id: Joi.string().required(),
+  newId: Joi.string(),
```
zcei: I think this needs to be required as well, as we don't use the whole object as a schema, but pick by property name and use the value as the schema.

So basically `Joi.validate(value, Joi.string())` is called, which would allow for empty content, am I right?

test/integration/migration.spec.js
```diff
@@ -171,6 +172,31 @@ describe('the migration', function () {
     ]);
   }));
 
+  it('changes field IDs', co(function * () {
+    yield migrator(idChange);
+
+    const resultAfterModify = yield request({
+      method: 'GET',
+      url: `/content_types/dog`,
```
zcei: No need for the backticks here 🙂

test/unit/lib/migration-payloads/validation/payload-validation-id-change.spec.js
```diff
@@ -0,0 +1,225 @@
...
+            name: 'Title',
+            type: 'Symbol'
+          }]
+        }
+      ];
+
+      const chunks = migrationChunks(steps);
+      const plan = migrationPlan(chunks);
+      const payloads = migrationPayloads(plan, existingCts);
+      const errors = validatePayloads(payloads);
```
zcei: Just a random thought: to not always repeat the same LOCs, it would be nice if we write a helper that creates `makeRequest` stubs which resolves with the `existingCts`, then we only need to run the `migrationParser`:

```js
const makeRequest = resolveWith(existingCts)

try {
  createMigrationParser(makeRequest)(migrationFn)
} catch (error) {
  expect(error.details).to.eql({ ... })
}
```

lib/migration-steps/validation.js
jfr3000: It wouldn't allow for an empty string but it would allow no value at all. Thanks for the catch, updating..

test/integration/migration.spec.js
jfr3000: copy :spaghetti:

test/unit/lib/migration-payloads/validation/payload-validation-id-change.spec.js
jfr3000: for now I've extracted these function calls into a `validateMigration` helper

madtrick:
@jfr3000 have you tried doing something like:

```
ct.editField('foo').id('bar')
ct.editField('bar').name('something something')
```

Looking at the code at https://github.com/contentful/migration-cli/blob/master/lib/migration-chunks/validation/field.js#L34 I wonder if the migration will break on that case since the library would think that the field does not exist.

jfr3000:
@madtrick thanks for pointing that out, totally overlooked that case. will add a test and update.

lib/migration-steps/index.js
```diff
@@ -92,6 +92,20 @@ class ContentType extends DispatchProxy {
 
     this.dispatch(actionCreators.field.delete(this.id, this.instanceId, id, fieldInstanceId, callsite));
   }
+
+  changeFieldId (oldId, newId) {
+    const callsite = getFirstExternalCaller();
+    const fieldInstanceId = this.fieldInstanceIds.getNew(oldId);
+    this.dispatch(actionCreators.field.update(
```
TimBeyer: Shouldn't we emit a new action type 'field rename'?
That way things would be very explicit and there's no special treatment based on a certain property (`newId`) being updated.

The actual payload change would then be done in a way similar to how we transform the field deletion into two updates.

test/unit/lib/migration-payloads/validation/validateMigration-helper.js
```diff
@@ -0,0 +1,15 @@
+const Bluebird = require('bluebird');
```
TimBeyer: The name of this module is a it odd :)
Maybe let's call it `validate-payloads.js` or something like that.
I think we could leave out the `-helper` part since all our actual tests are using `.spec.js` so there shouldn't be much confusion.

lib/migration-chunks/validation/field.js
```diff
@@ -166,17 +210,18 @@ module.exports = function (chunks, contentTypes = []) {
...
   const recentlyMoved = {};
+  const changedFieldIds = {};
 
   for (const chunk of chunks) {
     const contentTypeId = chunk[0].payload.contentTypeId;
     const contentTypeExists = Boolean(contentTypes.find((ct) => ct.sys.id === contentTypeId));
     const fieldSet = contentTypeFields[contentTypeId] || new Set();
     const fieldRemovals = recentlyRemoved[contentTypeId] || new Set();
     const fieldMovements = recentlyMoved[contentTypeId] || new Set();
+    const fieldIdChanges = changedFieldIds[contentTypeId] || {};
```
zcei: Let's use a `new Map()` here, so we call `.delete(key)` methods everywhere instead of `delete foo[key]`  expression.
Same for `.add(key)` instead of `foo[key] = ...`

lib/migration-chunks/validation/field.js
```diff
@@ -40,6 +40,10 @@ function fieldHasBeenMoved (validationContext, movementKey) {
   return validationContext.fieldMovements.has(movementKey);
 }
 
+function maybeGetChangedId (validationContext, id) {
+  return validationContext.fieldIdChanges && validationContext.fieldIdChanges[id];
```
zcei: This will never be undefined, as we create the `Map` (currently: object) before passing in the context

lib/migration-chunks/validation/field.js
```diff
@@ -153,9 +168,38 @@ const checks = {
...
+      errors.push(invalidActionError(
+        fieldErrors.changeId.ID_ALREADY_EXISTS(
+          oldId,
+          newId,
+          step.payload.contentTypeId
+        ),
+        step
+      ));
+    }
+    // if the id is switched back to something changed before, we forget that id
```
zcei: Shouldn't this add an error, as we rename the same field twice in one migration?

lib/migration-payloads/index.js
```diff
@@ -21,7 +21,7 @@ function buildFields (result, actions) {
       ));
     }
 
-    if (action.type === 'field/update') {
+    if (action.type === 'field/update' || 'field/rename') {
```
zcei:
```js
if (action.type === 'field/update' || action.type === 'field/rename')
```

lib/migration-payloads/index.js
```diff
@@ -76,6 +76,20 @@ function builder (chunk, contentType) {
   return result;
 };
 
+function maybeUpdateFieldIds (payload, fieldIdChanges) {
+  const fields = payload.payload.fields;
+  if (fields.some((field) => field.newId)) {
+    fieldIdChanges[payload.meta.contentTypeId] = _.cloneDeep(payload);
+  }
+  return fields.map((field) => {
```
zcei: Either `.map` & copy the object, or `for-of` with mutation.

test/integration/migration.spec.js
```diff
@@ -171,19 +172,60 @@ describe('the migration', function () {
...
+      method: 'GET',
+      url: '/content_types/dog',
+      headers: {
+        'X-Contentful-Beta-Dev-Spaces': 1
+      }
+    });
+
+    expect(resultAfterModify.fields).to.eql([
+      {
+        id: 'goodboys',
```
zcei: Shouldn't this be `aDifferentId`?

test/integration/migration.spec.js
```diff
@@ -171,19 +172,60 @@ describe('the migration', function () {
...
+        validations: []
+      }
+    ]);
+
+    let thisShouldNotExist;
+
+    try {
+      thisShouldNotExist = yield request({
+        method: 'GET',
+        url: '/content_types/aDifferentId',
```
zcei: You cannot rename a Content Type, so this can be dropped.

test/unit/lib/migration-chunks/validation/field/change-id.spec.js
```diff
@@ -0,0 +1,219 @@
...
+                'name': 'new Title'
+              }
+            }
+          }
+        }
+      }
+    ]);
+  }));
+
+  it('does not return errors when switching ids for several fields and several times on the same field', Bluebird.coroutine(function * () {
```
zcei: I think renaming the same field back & forth in the same migration shouldn't be possible, but renaming `title -> oldTitle` & `someOtherField -> title` in the same migration should work.

test/unit/lib/migration-chunks/validation/field/change-id.spec.js
```diff
@@ -0,0 +1,219 @@
...
+      }
+    ]);
+  }));
+
+  it('does return an error when switching the id of 2 fields to the same id', Bluebird.coroutine(function * () {
+    const steps = yield migrationSteps(function up (migration) {
+      const book = migration.editContentType('book');
+      book.changeFieldId('title', 'newTitle');
+      book.changeFieldId('pages', 'title');
+      book.changeFieldId('newTitle', 'title');
```
zcei: Maybe not rename `title -> newTitle` first, as it's then a "circular renaming" again.

lib/migration-chunks/validation/field.js
jfr3000: No, it could be a different field on the same content type that is getting the id that another field previously had (like here https://github.com/contentful/migration-cli/blob/5ebc8db380f937ce76f8e9f593bff73775fabec5/test/unit/lib/migration-chunks/validation/field/change-id.spec.js#L59)

lib/migration-payloads/index.js
jfr3000: :scream:

lib/migration-chunks/validation/field.js
zcei: I think I see your point, but in this example you actually do rename the same field twice 😆 

So if we rename `a -> b` and `c -> a` we forget about the `a -> b` renaming?

test/integration/migration.spec.js
jfr3000: No, because in this example I switched the id back after changing the name. But it is a bit contrived, I'll change the test to the more normal path.

test/integration/migration.spec.js
jfr3000: Um yes, I meant to get the old field here, not a content type. Removing it.

test/integration/migration.spec.js
zcei: Ah I missed the last line where it's renamed back (actually that makes a point for not allowing to rename the same field twice in a "circular" manner)

lib/migration-chunks/validation/field.js
jfr3000: Yes. Because if someone then tries to update `a`, that is now possible and we no longer want to tell them "hey, you can't do that because a has been renamed"

PR #15, opened 2017-09-27
Add deleteContentType method

jfr3000:
This adds a `.deleteContentType` method to the migration object, allowing users to delete content types.
A check for whether there still are any entries associated with the content type is performed.

A contentType/delete action does not have a payload or fields, requiring a bit of special treatment throughout the process.

Also part of this PR are two refactors - one removing the content-type-id module, whose functionality wasn't used by anything, as well as a refactoring the chunks content type validation, structuring it a bit.


lib/migration-chunks/validation/errors.js
```diff
@@ -83,6 +83,20 @@ module.exports = {
...
+    },
+    delete: {
+      CONTENT_TYPE_DOES_NOT_EXIST: (id) => {
+        return `You cannot delete content type "${id}" because it does not exist.`;
+      },
+      CONTENT_TYPE_ALREADY_DELETED: (id) => {
+        return `Content type with id "${id}" cannot be deleted more than once.`;
+      },
+      EDIT_AFTER_DELETE: (id) => {
+        return `Content type with id "${id}" cannot edited because it was deleted before.`;
```
zcei: > cannot **be** edited

PR #17, opened 2017-09-27
fix(fields): Before id change, check if field exists

jfr3000:
When validating a field ID change, we should perform all the basic checks that are done on a field update as well (such as checking if the field exists at all). 
That's why the 'field/update' check error messages should be worded more generally ("edit field" instead of "set property on field") so they work for edits of all kinds.

### DIFF
--- a/test/unit/lib/migration-chunks/validation/field/change-id.spec.js
+++ b/test/unit/lib/migration-chunks/validation/field/change-id.spec.js
@@ -182,7 +182,7 @@ describe('chunks validation when changing field ids', () => {
     ]);
   }));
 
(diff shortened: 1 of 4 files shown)


test/unit/lib/migration-chunks/validation/field/change-id.spec.js
```diff
@@ -258,4 +258,43 @@ describe('chunks validation when changing field ids', () => {
       }
     ]);
   }));
+
+  it('when changing the ID of a field but that field does not exist', Bluebird.coroutine(function * () {
```
zcei: maybe smth like `it('fails when changing the ID of a field that does not exist')`, to preserve the "fluent readability"

PR #18, opened 2017-09-28
fix: Do not allow setting of same property multiple times

jfr3000:
Previously, users could set a property on the same entity multiple times. This got collapsed into the correct payload but was showing the duplicates when rendering the plan. 

This PR introduces duplicate props validation for chunks, meaning props cannot be set multiple times within the same chunk (but setting them several times in separate chunks is ok).

The validation is happening after content type and fields validation, because it needs to look at a chunk including all its steps and because it makes sense to perform those basic validations first.

### DIFF
--- a/lib/migration-chunks/validation/duplicate-props.js
+++ b/lib/migration-chunks/validation/duplicate-props.js
@@ -1,0 +1,63 @@
+const errorMessages = require('./errors');
+
+const isSameEntity = (step, otherStep) => {
+  return (
+    step.payload.contentTypeId === otherStep.payload.contentTypeId &&
+    step.payload.fieldId === otherStep.payload.fieldId
+  );
+};
+
+const getErrorMsgParams = (step) => {
+  const type = step.type.split('/')[0];
+  const params = {
+    field: {
+      type: 'field',
+      id: step.payload.fieldId
+    },
+    contentType: {
+      type: 'content type',
+      id: step.payload.contentTypeId
+    }
+  };
+  return params[type];
+};
+
+const getDuplicateProps = (step, chunk, stepIndex) => {
+  return Object.keys(step.payload.props).reduce((duplicateProps, prop) => {
+    const isDuplicate = chunk.slice(0, stepIndex).some((stepBefore) => {
+      return stepBefore.payload.props &&
+        isSameEntity(step, stepBefore) &&
+        stepBefore.payload.props.hasOwnProperty(prop);
+    });
+    if (isDuplicate) {
+      duplicateProps.push(prop);
+    }
+    return duplicateProps;
+  }, []);
+};
+
+const getErrorsForChunk = (chunk) => {
+  return chunk.reduce((chunkErrors, step, stepIndex) => {
+    if (step.payload.props) {
+      const duplicateProps = getDuplicateProps(step, chunk, stepIndex);
+      const errors = duplicateProps.map((duplicateProp) => {
+        const { type, id } = getErrorMsgParams(step);
+        return {
+          type: 'InvalidAction',
+          message: errorMessages.DUPLICATE_PROP(duplicateProp, type, id),
+          details: { step }
+        };
+      });
+      chunkErrors = chunkErrors.concat(errors);
+    }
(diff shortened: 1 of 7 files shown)


lib/migration-chunks/validation/index.js
```diff
@@ -27,6 +28,11 @@ module.exports = function (chunks, contentTypes) {
...
 
   const allCTs = contentTypes.concat(createdCTs);
+  const fieldErrors = fieldValidations(chunks, allCTs);
 
-  return fieldValidations(chunks, allCTs);
+  if (fieldErrors.length > 0) {
+    return fieldErrors;
+  }
+
+  return checkForDuplicatePropsErrors(chunks);
```
TimBeyer: Wondering if we shouldn't merge this together with the previous errors.

lib/migration-chunks/validation/duplicate-props.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+  };
+  return params[type];
+};
+
+module.exports = (chunks) => {
+  return chunks.reduce((errors, chunk) => {
+    const chunkErrors = chunk.reduce((chunkErrors, step, i) => {
+      if (hasProps(step)) {
+        const thisStepsProp = Object.keys(step.payload.props)[0];
```
TimBeyer: I guess in theory (even though in our implementation we only ever emit one prop) we should check all of them.

zcei: Will this always be just one prop? 🤔

lib/migration-chunks/validation/duplicate-props.js
```diff
@@ -0,0 +1,52 @@
...
+        const duplicate = chunk.slice(0, i).find((stepBefore) => {
```
zcei: Can we pull this into `hasDuplicate(previousSteps, step)` to remove one nesting level?

lib/migration-chunks/validation/duplicate-props.js
```diff
@@ -0,0 +1,52 @@
...
+      type: 'content type',
+      id: step.payload.contentTypeId
+    }
+  };
+  return params[type];
+};
+
+module.exports = (chunks) => {
+  return chunks.reduce((errors, chunk) => {
+    const chunkErrors = chunk.reduce((chunkErrors, step, i) => {
```
zcei: Not too sure about this, but it took me a minute to grasp what the double-nested `reduce` does.
Wondering whether it makes it more readable, if we pull the reducer into its own fn, but maybe it just makes you jump between the fns more.

lib/migration-chunks/validation/duplicate-props.js
```diff
@@ -0,0 +1,52 @@
+const errorMessages = require('./errors');
+
+const isSameEntity = (step, otherStep) => {
+  return (
+    step.payload.contentTypeId === otherStep.payload.contentTypeId &&
+    step.payload.fieldId === otherStep.payload.fieldId
+  );
+};
+
+const hasProps = (step) => step.payload && step.payload.props;
```
zcei: By our convention (based on the FSA) there should always be a payload - just maybe not as an object. So maybe a `typeof step.payload === 'object'` would make it clearer, that we only want actions in the format of `{ payload: { foo: 'bar' } }`

test/unit/lib/migration-chunks/validation/content-type.spec.js
```diff
@@ -347,4 +347,39 @@ describe('content type plan validation', function () {
       ]);
     }));
   });
+
+  describe('when setting the same prop more than once in one chunk', function () {
+    it('returns an error', Bluebird.coroutine(function * () {
+      const steps = yield migrationSteps(function up (migration) {
+        const person = migration.createContentType('person').name('Person');
+        person.name('Person McPersonface');
```
zcei: 🤣

lib/migration-chunks/validation/duplicate-props.js
zcei: Mentioned by @TimBeyer already in the meantime

PR #19, opened 2017-09-29
chore: Include linter in npm test command

jfr3000:


TimBeyer:
Oh just realized I'm a bit late to the party

jfr3000:
@Tim, that's true. Will open one more.

Issue #20, opened 2017-10-02
Validation with custom message


jody-zeitler:
If I define a field with an asset validation

```js
module.exports = function (migration) {
	const application = migration.createContentType('application')
		.name('Application')
		.displayField('displayName');

	application.createField('avatar')
		.name('Avatar')
		.type('Link')
		.linkType('Asset')
		.localized(true)
		.validations([
			{
				"linkMimetypeGroup": [
					"image"
				]
			},
			{
				"assetFileSize": {
					"max": 19922944
				},
				"message": "Error Message: The file size of your image is to large. The max file size is 19MB."
			}
		]);
};
```

plan validation fails with

```shell
Validation failed

The following migration has been planned but cannot be run because it contains errors

Update Content Type application
  - name: Application
  - displayField: displayName

  Update field avatar
    - name: "Avatar"
    - type: "Link"
    - linkType: "Asset"
    - localized: true
    - validations: [{"linkMimetypeGroup":["image"]},{"assetFileSize":{"max":19922944},"message":"Error Message: The file size of your image is to large. The max file size is 19MB."}]


Error: A field can't have "assetFileSize" as a validation.

🚨  Migration unsuccessful
```

If I remove the "message" property, the migration plan is accepted and "assetFileSize" appears correctly within the target space.

zcei:
Hej @jody-zeitler,

good catch, the `message` property totally slipped through!

I'm working on a fix for both (allowing `message` key & giving a more meaningful error message) and will release a semver patch version afterwards.

PR #22, opened 2017-10-04
Show invalid combination of validation props

zcei:
As mentioned in #20, currently the validation message is not very meaningful when you have an additional un-allowed property in an otherwise valid validation object.

### DIFF
--- a/lib/migration-payloads/validation/errors.js
+++ b/lib/migration-payloads/validation/errors.js
@@ -50,6 +50,16 @@ const errors = {
       INVALID_VALIDATION_PROPERTY: (propName) => {
         return `A field can't have "${propName}" as a validation.`;
       },
+      INVALID_VALIDATION_PROPERTY_COMBINATION: (keys) => {
+        const wrappedKeys = keys.map((key) => `"${key}"`);
+        const list = wrappedKeys.slice(0, -1);
+        const [last] = wrappedKeys.slice(-1);
+
+        const commaSeparatedKeys = list.join(', ');
+        const formattedCombination = `${commaSeparatedKeys} and ${last}`;
+
+        return `A field can't have the combination ${formattedCombination} as a validation.`;
+      },
       INVALID_VALIDATION_PARAMETER: (propName, expectedType, actualType) => {
         return `"${propName}" validation expected to be "${expectedType}", but got "${actualType}"`;
       }
--- a/lib/migration-payloads/validation/schema-validation.js
+++ b/lib/migration-payloads/validation/schema-validation.js
@@ -132,6 +132,45 @@ const validateContentType = function (payload) {
   });
 };
 
+const unknownCombinationError = function ({ path, keys }) {
+  const type = 'object.unknownCombination';
+  const message = `"${keys.join()}" is not allowed`;
+  const context = { keys };
+
+  return { message, path, type, context };
+};
+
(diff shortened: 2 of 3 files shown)


lib/migration-payloads/validation/errors.js
```diff
@@ -50,6 +50,10 @@ const errors = {
       INVALID_VALIDATION_PROPERTY: (propName) => {
         return `A field can't have "${propName}" as a validation.`;
       },
+      INVALID_VALIDATION_PROPERTY_COMBINATION: (keys) => {
+        const formattedCombination = keys.map((key) => `"${key}"`).join(' & ');
```
TimBeyer: I would try to find a way not to use the ampersand.

lib/migration-payloads/validation/schema-validation.js
```diff
@@ -132,6 +132,38 @@ const validateContentType = function (payload) {
...
 
+const unknownCombinationError = function ({ path, keys }) {
+  const type = 'object.unknownCombination';
+  const message = `"${keys.join()}" is not allowed`;
+  const context = { keys };
+
+  return { message, path, type, context };
+};
+
+const combineErrors = function (fieldValidationsErrors) {
```
TimBeyer: Maybe you could add a comment that shows the structure we're dealing with for every step of transformations so it's more easily understandable.

lib/migration-payloads/validation/errors.js
```diff
@@ -50,6 +50,10 @@ const errors = {
...
+        return `A field can't have the combination ${formattedCombination} as a validation.`;
```
jfr3000: I'd phrase this more as "x is not a valid property for a validation" The combination thing is a bit confusing.

test/unit/lib/migration-payloads/validation/payload-validation-field-validations.spec.js
```diff
@@ -63,6 +63,32 @@ describe('payload validation', function () {
...
+      const chunks = migrationChunks(steps);
+      const plan = migrationPlan(chunks);
+      const payloads = migrationPayloads(plan);
+      const errors = validatePayloads(payloads);
+
+      expect(errors).to.eql([
+        [
+          {
+            type: 'InvalidPayload',
+            message: `A field can't have the combination "size" & "somethingUnknown" as a validation.`
```
TimBeyer: I find the error message somewhat misleading.  
I guess the problem is that our validations are just simple keys with the actual validation config being the value.  
So the error is more of a 'A validation can only have one key' error or something like that. 

I think ideally if there is one valid and `n` invalid properties we'd say that the valid validation cannot have any other keys in there, if all of them are invalid we just list them all as invalid, and if there are multiple valid ones we mention that they need to be separated.

At the same time it's all more of a nice to have, so I don't insist on it :)

zcei: > A validation can only have one key

But a validation can have multiple keys, see #23.

TimBeyer: Oh, right. Damn :(

zcei:
@TimBeyer addressed your feedback.

I think for now this should be sufficient - I'm playing with a different way of handling the additional properties, which would then allow us to access the list of allowed properties, so we could filter them and only display the *really* unknown ones.

PR #24, opened 2017-10-04
chore: Run linter after tests

jfr3000:


zcei:
I'm not too sure I like this. Then you're all set and everything went through, only to see there's a semicolon missing somewhere and then the whole cycle starts again? 🤔

PR #27, opened 2017-10-06
chore: refactor slicers

jfr3000:
Extracts the slicer logic for field/delete and field/rename. The migration-chunks test pass, but the unit tests overall don't because of WIP on the feature branch.

The `intoSeparateChunk` slicer now uses a function that returns true once, the version before was unnecessarily complicated.


jfr3000:
Just realized that the stuff I refactored is already on master, I'll open a new PR against master.

PR #28, opened 2017-10-06
chore(chunks): refactor slicers

jfr3000:
Extracts the slicer logic for field/delete and field/rename.

The `intoSeparateChunk` slicer now uses a function that returns true once, the version before was unnecessarily complicated.


lib/migration-chunks/slicers/into-separate-chunk.js
```diff
@@ -0,0 +1,24 @@
...
+    return ret;
+  };
+};
+
+module.exports = (itemType) => {
+  const isItemType = (item) => item.type === itemType;
+  return createPredicateSlicer((list) => {
+    if (isItemType(list[0])) {
+      return returnTrueOnce();
+    }
```
madtrick: nit and a bit OT to this PR but could we fix the eslint rules to require an empty line after `if` blocks?

lib/migration-chunks/slicers/into-separate-chunk.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const createPredicateSlicer = require('./predicate');
+
+const returnTrueOnce = function () {
```
madtrick: nit: this seems a bit restrictive to be used by other slicers in the future and it also has implicit knowledge of how the `field-id-change` and `field-delete` chunks are formed.

lib/migration-chunks/slicers/into-separate-chunk.js
jfr3000: I'm not sure this is something that needs to be regulated by a linter rule but we can discuss it in the team I guess?

lib/migration-chunks/slicers/into-separate-chunk.js
jfr3000: Not sure I understand, this slicer will put an item of a given type into its own chunk and will work for any type of item. Can you explain where the implicit knowledge is?

lib/migration-chunks/slicers/into-separate-chunk.js
madtrick: we can discuss it in the team and if we agree it's something we want we should put in on the linter. Otherwise we'll have this conversation next time someone forgets to add the whiteline : )

lib/migration-chunks/slicers/into-separate-chunk.js
madtrick: By implicit knowledge I meant the fact that `field-id-change` and `field-delete` expect to be called only once since field deletion and id change have to always go on independent chunks. But this not explicitly stated anywhere. Also, if we later add another slicer which needs to group by more than one step this won't work.

jfr3000: I see... To me, the module name `into-separate-chunk` along with the way it's called
```
const sliceByFieldDelete = require('./slicers/into-separate-chunk')('field/delete');
```
is explicit enough. I could add a comment for `returnTrueOnce` that explains that?
As for other slicer that need to group differently: Sure, the other slicers remain the same, and we can always add different ones as well. But if we encounter another case where a step of a type needs to be alone in a chunk, we can use this one.

PR #29, opened 2017-10-06
test(e2e): extend e2e tests by examples coverage

stupchiy:


test/end-to-end/client.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const Bluebird = require('bluebird');
+
+const path = require('path');
+const { createManagementClient } = require('../../bin/lib/contentful-client');
+
+const config = {
+  accessToken: process.env.CONTENTFUL_INTEGRATION_MANAGEMENT_TOKEN,
+  application: 'contentful.migration-cli.e2e-test/0.0.0'
```
TimBeyer: Nice to have a separate useragent thingy here. Maybe we could even `require` the `package.json` and use the version number from there.

test/end-to-end/client.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const client = createManagementClient(config);
+
+const makeRequest = function (spaceId, requestConfig) {
+  requestConfig.url = path.join(spaceId, requestConfig.url);
+  return client.rawRequest(requestConfig);
+};
+
+const waitForJobCompletion = Bluebird.coroutine(function * (makeRequest, spaceId, jobId) {
```
TimBeyer: Would be cool to also share this code with the integration tests since it's somewhat duplicated

stupchiy: good idea, will update

test/end-to-end/client.js
stupchiy: good idea, will update

PR #38, opened 2017-11-08
test(e2e): added basic tests for syntax errors validations

stupchiy:
Syntax errors validation tests for:
- undefined function 
- missing argument
- invalid token


test/end-to-end/payload.spec.js
```diff
@@ -0,0 +1,31 @@
...
+      .run(`--space-id ${SOURCE_TEST_SPACE} ./examples/10-payload-undefined.js`)
+      .expect(assert.payload.notDefined('buggybug'))
+      .end(done);
+  });
+
+  it('outputs the missing argument error', function (done) {
+    const script = '11-payload-corrupt.js';
+    cli()
+      .run(`--space-id ${SOURCE_TEST_SPACE} ./examples/${script}`)
+      .expect(assert.payload.syntaxError(script, 4, 'missing ) after argument list'))
```
zcei: Wow, we're that smart? Thought we just say it's wrong JS syntax 👍

jfr3000:
these are good, but I think @TimBeyer meant something different. payload-validation is the name of an internal module so that's confusing. what that that module validates are errors across a whole migration script, for example, when a content type is edited before it is created. all the things it checks for can be found here: https://github.com/contentful/migration-cli/blob/master/lib/migration-payloads/validation/errors.js

PR #47, opened 2017-11-10
Release pipeline

zcei:
In order to not distribute unnecessary stuff, we `.npmignore` a good chunk of stuff.
Note that `.DS_Store` and alike are ignored by default ([full list](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)).

It then wipes the `built` folder and triggers a rebuild - this is hooked to the `prepare` lifecycle, so that it gets executed after `npm install` and before `npm publish`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,9 @@
     "url": "git://github.com/contentful/migration-cli.git"
   },
   "scripts": {
-    "build": "tsc -p .",
+    "clean": "rimraf built",
+    "build": "npm run clean && tsc -p .",
+    "prepare": "npm run build",
     "test": "npm run build && npm run test-unit && npm run test-integration && npm run test-e2e && npm run lint",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -113,5 +116,8 @@
   },
   "bin": {
     "contentful-migration": "./bin/contentful-migration"
+  },
+  "engines": {
+    "node": ">=8.0.0"
```
madtrick: After replacing `Object.value` with `_,values` we could target also node v6. Is there anything else preventing us from doing so?

zcei: Currently we have set the compile target to `es2017`, mainly for cleaner stack traces.
When not supporting Node 6, we get rid of all the `Generator` lines in the trace.

TimBeyer: We do not need to turn async/await into generators then

PR #48, opened 2017-11-13
Add yes flag to skip confirmation

zcei:


src/bin/cli.ts
```diff
@@ -128,11 +133,19 @@ const run = async function () {
     }
   })
 
-  const answers = await inquirer.prompt([{
-    type: 'confirm',
-    message: 'Do you want to apply the migration',
-    name: 'applyMigration'
-  }])
+  const confirm = async function ({ skipConfirmation = false }) {
```
madtrick: Nit: if there's already a default set for the `yes` option in `yargs`, why also set it here?

zcei: I was initially thinking to reshuffle a bit more, so that the `run` function would accept all the inputs that currently reside in `argv`, which would allow for a "library usage", but then considered that out-of-scope.

Leftover from there, but I think it:
a) doesn't hurt
b) makes sure we won't forget a default when restructuring
c) allows for TypeScript type inference

examples/empty-migration.js
```diff
@@ -0,0 +1 @@
+module.exports = function () {};
```
TimBeyer: Maybe we should stop putting all these e2e test related files into the examples directory.  
This is definitely not a useful example for anyone ;)

src/bin/cli.ts
madtrick: d) can can cause unexpected behaviour if we were to change the default in `yargs` (or in the function definition) and still got the same old result 😉

zcei: Well, if you change the default in `yargs`, there will be a value set, thus overwriting the default of the function 🙃

examples/empty-migration.js
zcei: Was thinking the same, but put them there to not start a new "structure paradigm" for such a simple change.

stupchiy:
nice, I like the tests presence 👍

src/bin/cli.ts
madtrick: I still think is confusing to have it in two places besides if someone changes the default in one of the places has to remember to change in the other place too. Maybe at least pull the default to a constant so it's more visible that is not a magic boolean.

zcei: Just went with a type annotation now, and leaving the rest up to the consumer (i.e. yargs)

PR #49, opened 2017-11-16
Transform content

TimBeyer:


src/lib/fetcher.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export default class Fetcher {
+  private makeRequest: any
+
+  constructor (makeRequest) {
+    this.makeRequest = makeRequest
+  }
+
+  async getEntriesInIntents (intentList: IntentList) {
+    const ids: Array<string> = _.uniq(intentList.getIntents().map((intent) => intent.getContentTypeId()))
```
TimBeyer: I don't think it makes a huge difference but I prefer `const ids: string[]` instead of `Array<string>`

src/lib/fetcher.ts
```diff
@@ -0,0 +1,81 @@
...
+      .filter((intent) => intent.isContentTypeDelete())
+      .map((intent) => intent.getContentTypeId())
+    )
+
+    if (deletedCtIds.size === 0) {
+      return contentTypes
+    }
+
+    const self = this
+    return Bluebird.map(contentTypes, async function (ct: any): Promise<any> {
```
TimBeyer: This function returns `Promise<ContentType>` and not `Promise<any>`

src/lib/action/entry-derive.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+        await api.saveEntry(targetEntry.id)
+        if (this.shouldPublish) {
+          await api.publishEntry(targetEntry.id)
+        }
+      }
+
+      // TODO: `setFieldForLocale` does not create the path
+      // to the field object and assigns directly.
+      // So if the field never existed before it fails
```
zcei: This sounds like a quite legit use-case and creating the path shouldn't be too hard - let's fix this before merge?

src/lib/action/entry-transform.ts
```diff
@@ -1,31 +1,52 @@
...
   private contentTypeId: string
   private fromFields: string[]
-  private toFields: string[]
-  private transform: Function
+  // private toFields: string[]
+  private transformEntryForLocale: Function
+  private shouldPublish: boolean
 
-  constructor (contentTypeId: string, fromFields: string[], toFields: string[], transform: Function) {
+  // constructor (contentTypeId: string, fromFields: string[], toFields: string[], transformation: Function) {
```
zcei: Is this the "next generation constructor" or can we get rid of this?

src/lib/fetcher.ts
```diff
@@ -0,0 +1,81 @@
...
+    const response = await this.makeRequest({
+      method: 'GET',
+      url: `/entries?sys.contentType.sys.id[in]=${ids.join(',')}`
+    })
+
+    return response.items
+  }
+
+  async getContentTypesInChunks (intentList: IntentList) {
+    const ids: Array<string> = _.uniq(intentList.getIntents().map((intent) => intent.getContentTypeId()))
```
zcei: Here again `string[]` for consistency

src/lib/fetcher.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    const response = await this.makeRequest({
+      method: 'GET',
+      url: `/entries?sys.contentType.sys.id[in]=${ids.join(',')}`
+    })
+
+    return response.items
+  }
+
+  async getContentTypesInChunks (intentList: IntentList) {
```
zcei: return type is `APIContentType[]`, better to add it explicitly, so that it won't change unintended

src/lib/intent/entry-transform.ts
```diff
@@ -8,12 +8,11 @@ export default class EntryTransformIntent extends Intent {
     return true
   }
 
-  groupsWith (other: Intent): boolean {
-    const sameContentType = other.getContentTypeId() === this.getContentTypeId()
-    return other.isContentTransform() && sameContentType
```
zcei: So we decided to never group content transforms together? What changed our mind?

src/lib/migration-chunks/validation/errors.ts
```diff
@@ -9,7 +9,10 @@ interface ErrorCreators {
   contentType: {
     [groupIdentifier: string]: ErrorGroup
   },
-  generic: ErrorGroup
+  generic: ErrorGroup,
+  entry: {
+    [groupIdentifier: string]: ErrorGroup
+  }
```
zcei: Small nit, maybe have `generic` as last item

src/lib/migration-chunks/validation/index.ts
```diff
@@ -26,6 +26,7 @@ function validateChunks (intentList: IntentList, contentTypes: ContentType[]): v
   if (fieldErrors.length > 0) {
     throw new ErrorCollection(fieldErrors, {payloadValidationError: true})
   }
+
```
zcei: ESLint would have complained 😛

src/lib/migration-parser.ts
```diff
@@ -57,17 +58,19 @@ const createMigrationParser = function (makeRequest): (migrationCreator: () => a
...
+    const locales = await fetcher.getLocalesForSpace()
+
+    const api = new OfflineAPI(existingCts, entries, locales)
 
     await intentList.compressed().applyTo(api)
 
     const batches = await api.getRequestBatches()
 
-    return batches
+    return { batches }
```
zcei: This was mainly introduced for the `errorCollector` that we don't have anymore? Should we revert the interface to return only batches?

src/lib/offline-api/index.ts
```diff
@@ -321,6 +359,10 @@ class OfflineAPI {
     return this.requestBatches
   }
 
+  public recordTransformError (error) {
```
zcei: Depending on the use-case this might turn async later as well - or we have to always store errors in-memory

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
 
 Deletes the content type with the provided id and returns `undefined`. Note that the content type must not have any entries.
 
+#### `transformEntries(config)`
+
+For the given content type, transforms all entries according to the user-provided `transformEntryForLocale` function. This function will be called for every entry and every locale with the `from` fields and the current locale. It is expected to return an object with the desired target fields and the value OR undefined, in which case this entry locale will be skipped. If `transformEntryForLocale` does return a value, it will be written to the current locale.
```
madtrick: missing backticks around `undefined`

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+Please also refer to [this example](./examples/12-transform-content.js).
+
+**`config : Object`** – Content transformation definition, with the following properties:
+- **`contentType : string`** _(required)_ – Content type ID
+- **`from : array`** _(required)_ – Array of the source field IDs
+- **`to : array`** _(required)_ – Array of the target field IDs
+- **`transformEntryForLocale : function (fields, locale): object`** _(required)_ – Transformation function to be applied.
```
madtrick: We could maybe add here a description of the structure of fields and locale. Something like:

```markdown
- **`transformEntryForLocale : function (fields, locale): object`** _(required)_ – Transformation function to be applied.
    - `fields` is an object containing each of the `from` fields. Each field will contain their current localized values (i.e. `from == {myField: {'en-US': 'my field value'}}`)
    - `locale` one of the locales in the space being transformed
```

README.md
madtrick: I would also change the line `This function will be called for every entry and every locale with the `from` fields and the current locale.` to `For each entry, the cli will call this function once per locale in the space with ...`

madtrick: `For the given content type, transforms all entries...` maybe add an `its` --> `For the given content type, transforms all its entries...`

madtrick: Something else that I just thought: instead of  `... the desired target fields and the value OR undefined, in which case this entry locale will be skipped` I would write something like `... the desired target fields and the value OR undefined, in which case the values of the entry for the current locale will be left untouched.

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`shouldPublish : boolean`** _(optional)_ – Publish entries after transform (default _true_)
+
+#### `deriveLinkedEntries(config)`
+
+For the given content type, derives new entries from existing ones and sets up the new ones as a reference on the existing one according to the user-provided `deriveEntryForLocale` function. This function will be called for every entry and every locale with the `from` fields and the current locale. It is expected to return an object with the desired target fields and the value OR undefined, in which case this entry locale will be skipped.
```
madtrick: Another suggestion:

`For each entry of the given content type, derives a new entry and sets up a reference to it. The content of the new entry is generated by the user provided 'deriveEntryForLocale` function`

Some of the comments I made for `transformEntryForLocale` also apply to this function:

- `undefined`
- `skipped` vs `untouched`
- Clarification of `fields` and `locale` in `deriveEntryForLocale`

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`to : array`** _(required)_ – Array of the target field IDs
+- **`transformEntryForLocale : function (fields, locale): object`** _(required)_ – Transformation function to be applied.
+- **`shouldPublish : boolean`** _(optional)_ – Publish entries after transform (default _true_)
+
+#### `deriveLinkedEntries(config)`
+
+For the given content type, derives new entries from existing ones and sets up the new ones as a reference on the existing one according to the user-provided `deriveEntryForLocale` function. This function will be called for every entry and every locale with the `from` fields and the current locale. It is expected to return an object with the desired target fields and the value OR undefined, in which case this entry locale will be skipped. 
+Please also refer to [this example](./examples/15-derive-entry.js).
+
+**`config : Object`** – Content transformation definition, with the following properties:
```
madtrick: `Content **derivation** definition`?

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`contentType : string`** _(required)_ – Source entry content type ID
```
madtrick: I find that `entry` confusing. Couldn't we remove it and just use `Source content type ID`?

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`derivedContentType : string`** _(required)_ – Target entry content type ID
+- **`from : array`** _(required)_ – Array of the source field IDs
+- **`toReferenceField : string`** _(required)_ – Field ID on the source content type in which to create the reference
```
madtrick: I think it reads a bit nicer if we move the `field` word a bit: `Id of the field on the source ...`

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`derivedFields : array`** _(required)_ – Array of the field IDs on the target content type
+- **`identityKey: function (fields): string`** _(required)_ - Called once per entry to transform. Returns the ID used for the new entry, which is also used for deduplication so that multiple source entries can link to the same derived entry.
```
madtrick: Missing description of the *shape* and *values* in the argument `fields`

madtrick: For consistency I would change `Returns the ID used for the new entry` to `Returns the ID used for the derived entry` (`new` vs `derived`)

README.md
```diff
@@ -86,6 +86,35 @@ Uses the same options as [`createContentType`](#createcontenttypeid--string-opts
...
+- **`deriveEntryForLocale : function (fields, locale): object`** _(required)_ – Transformation function to be applied.
+- **`shouldPublish : boolean`** _(optional)_ – Publish entries after deriving (default _true_)
```
madtrick: What's the status of the derived entries? are they also published? or can the user control that?

examples/15-derive-entry.js
```diff
@@ -0,0 +1,35 @@
+module.exports = function (migration) {
```
madtrick: Missing comment describing the example

src/lib/action/entry-derive.ts
```diff
@@ -0,0 +1,92 @@
...
+import { OfflineAPI } from '../offline-api'
+import Entry from '../entities/entry'
+import * as _ from 'lodash'
+
+class EntryDeriveAction extends APIAction {
+  private contentTypeId: string
+  private fromFields: string[]
+  private referenceField: string
+  private derivedContentType: string
+  // TODO: We do not use this here at all
```
madtrick: To what does this comment refer? to `derivedFields`?

src/lib/intent/entry-transform.ts
jfr3000: I think at the time we weren't 100% sure if that could have any negative effects and decided to err on the side of caution.

src/lib/action/entry-transform.ts
madtrick: I think that for now we can just get rid of it

src/lib/action/entry-transform.ts
```diff
@@ -1,31 +1,52 @@
...
     for (const entry of entries) {
-      const inputs = this.fromFields.map((fieldId) => entry.fields[fieldId])
-      const outputs = await this.transform(inputs)
-      for (let i = 0; i < this.toFields.length; i++) {
-        entry.fields[this.toFields[i]] = outputs[i]
+      const inputs = _.pick(entry.fields, this.fromFields)
+      for (const locale of locales) {
+        let outputsForCurrentLocale
+        try {
+          outputsForCurrentLocale = await this.transformEntryForLocale(inputs, locale)
```
madtrick: Missing code to handle `undefined` the same way we say in the interface spec (right now it will crash at `Object.keys`)

src/lib/entities/entry.ts
```diff
@@ -30,6 +30,16 @@ class Entry {
     this._fields = fields
   }
 
+  setField (id: string, value: any) {
+    this._fields[id] = value
+  }
+
+  setFieldForLocale (id: string, locale: string, value) {
```
madtrick: nit for completion: missing type annotation for `value`

src/lib/intent-validator/content-transform.ts
```diff
@@ -0,0 +1,30 @@
...
+  // just the prop name
+  get propertyNameToValidate () {
+    return 'transformation'
+  }
+
+  get schema () {
+    return {
+      from: Joi.array().items(Joi.string()).required(),
+      to: Joi.array().items(Joi.string()).required(),
+      transformEntryForLocale: Joi.any(),
```
madtrick: Didn't @TimBeyer fix this to use the `flags: {_func: true}` hack?

madtrick: It's also missing the `required` property

src/lib/intent-validator/content-transform.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  // NOTE: this could be change to return the object to
+  // validate against the schema instead of returning
+  // just the prop name
+  get propertyNameToValidate () {
+    return 'transformation'
+  }
+
+  get schema () {
+    return {
```
madtrick: Aren't we missing here the `contentTypeId` property?

src/lib/intent/entry-transform.ts
madtrick: That we would have to explicitly state how to handle the case where one content transform publishes the entries but the other doesn't.

src/lib/interfaces/content-transform.ts
```diff
@@ -1,5 +1,6 @@
 export default interface ContentTransform {
   from: string[],
```
madtrick: Missing `contentTypeId`

src/lib/intent/index.ts
```diff
@@ -20,5 +21,6 @@ export {
   FieldCreateIntent as FieldCreate,
   ContentTypeCreateIntent as ContentTypeCreate,
   ContentTypeDeleteIntent as ContentTypeDelete,
-  ContentTypeUpdateIntent as ContentTypeUpdate
+  ContentTypeUpdateIntent as ContentTypeUpdate,
+  EntryDeriveIntent as EntryDerive
```
madtrick: Don't we have to also export the content transform intent? 🤔

src/lib/migration-chunks/validation/errors.ts
```diff
@@ -109,6 +112,52 @@ const errorCreators: ErrorCreators = {
...
+        return `You cannot transform entries for content type "${id}" because it does not exist.`
+      },
+      TRANSFORM_AFTER_CONTENT_TYPE_DELETE: (id) => {
+        return `You cannot transform entries for content type "${id}" because it was deleted.`
+      },
+      TRANSFORM_BEFORE_CONTENT_TYPE_CREATE: (id) => {
+        return `You cannot transform entries for content type "${id}" because it has not yet been created.`
+      }
+    },
+    deriveEntries: {
```
madtrick: question: do we run the same validations for both the source and target CTs?

README.md
zcei:
```diff
- cli
+ CLI
```

src/lib/migration-chunks/validation/errors.ts
jfr3000: yes!

src/lib/intent/index.ts
zcei: Isn't this `EntryTransformIntent as EntryTransform`?

PR #50, opened 2017-11-20
Fix docs

hlabas:
Minor fix to the README on the `identityKey` signature


TimBeyer:
Those coverage metrics reports are really really broken :(

PR #51, opened 2017-11-20
render errors with more details and write errors log

jfr3000:
This is a proposal to give more detail when server errors happen. It's hard to anticipate where in the error the most important info is, so this will display statusText, message, details and url. Additionally, the complete errors get written to a log. This is less pretty, but more helpful when trying to find out what actually went wrong. It will look like this:
image

### DIFF
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -1,5 +1,6 @@
 import * as path from 'path'
 import * as fs from 'fs'
+import * as util from 'util'
 import * as yargs from 'yargs'
 
 import chalk from 'chalk'
@@ -15,6 +16,8 @@ import renderPlan from './lib/plan-messages'
 import renderStepsErrors from './lib/steps-errors'
 import { RequestBatch } from '../lib/offline-api/index'
 import Fetcher from '../lib/fetcher'
+
+const appendFile = util.promisify(fs.appendFile)
 
 const argv = yargs
   .usage('Parses and runs a migration script on a Contentful space.\n\nUsage: contentful-migration [args] <path-to-script-file>\n\nScript: path to a migration script.')
@@ -126,6 +129,9 @@ const run = async function () {
     process.exit(1)
   }
 
+  const serverErrorsFile = path.join(process.cwd(), `errors-${Date.now()}`)
+  const serverErrorsWritten = []
+
   const tasks = batches.map((batch) => {
     return {
       title: batch.intent.toPlanMessage().heading,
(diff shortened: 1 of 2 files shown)


src/bin/cli.ts
```diff
@@ -112,20 +112,23 @@ const run = async function () {
     process.exit(1)
   }
 
-  const hasErrors = batches.some((batch) => batch.errors.length > 0 || batch.contentTransformErrors.length > 0)
+  const hasTransformErrors = batches.some((batch) => batch.errors.length > 0 || batch.contentTransformErrors.length > 0)
```
zcei: If you want to really check for `hasTransformErrors` only, we should split this up into two arrays, each filled by one predicate only

src/bin/cli.ts
```diff
@@ -112,20 +112,23 @@ const run = async function () {
...
   }
   renderPlan(batches)
 
-  if (hasErrors) {
+  if (hasTransformErrors) {
     console.log(chalk`🚨  {bold.red Migration unsuccessful}`)
     process.exit(1)
   }
 
+  const serverErrorsFile = path.join(process.cwd(), `errors-${Date.now()}`)
```
zcei: Lets put the migration name in between and end with a `.log` extension, like `errors-add-author-1511191373630.log` - WDYT?

src/bin/cli.ts
```diff
@@ -1,5 +1,5 @@
 import * as path from 'path'
-import * as fs from 'fs'
+import * as fs from 'mz/fs'
```
zcei: Now that we're Node 8 only, we could use `util.promisify` to get rid of one dependency.

src/bin/cli.ts
hlabas: +1 to that, no extension files are not always naturally associated with the right editors. I'd consider `.err` as well unless we plan to put anything else than errors in there (like a detailed log of the migration's execution, with elements like `console.log` calls for example.

src/bin/cli.ts
jfr3000: yes!

src/bin/cli.ts
jfr3000: ah, I wanted to make the difference to the server errors below clearer but overlooked that this is actually different types of errors. reversing this one.

src/bin/cli.ts
jfr3000: @zcei what do you mean by migration name, where does that come from?

as for the file extension, I'm not sure. giving it one feels not very Unix... what do @TimBeyer and @madtrick think?

zcei: @jfr3000 that's the migrations' file name portion `path.basename(argv.filePath, '.js')`

TimBeyer: I think it might make sense to give it an extension.  
What are we currently logging? Maybe we can log as http://specs.okfnlabs.org/ndjson/ and give it the ending `.ndjson`

TimBeyer:
What do we do about transform errors?

jfr3000:
That'll be a separate PR but I'm thinking to log those into a file as well. This one is just about the server errors.

src/bin/cli.ts
jfr3000: We're logging the complete error objects, so `.ndjson` could make sense.

jfr3000:
I'll open another PR that's for both server and runtime errors, addressing the change requests everyone made.

PR #52, opened 2017-11-21
Add rendering and logging for runtime and server errors

jfr3000:
And another round :)
In addition to server errors, runtime errors that happen during a transform or derive are rendered and logged.

The renderers for validation and runtime errors have been extracted into a module, as well as the logfile module.

For runtime errors, it'll look like this
image

And for server errors like this
image


src/bin/cli.ts
```diff
@@ -170,8 +176,10 @@ const run = async function () {
       return successfulMigration
     } catch (err) {
       console.log(chalk`🚨  {bold.red Migration unsuccessful}`)
-      console.log(chalk`{red ${err.message}}`)
-      console.log(err.errors.map((err) => chalk`{red ${err.message}}\n`))
+      console.log(chalk`{red ${err.message}}\n`)
+      err.errors.forEach((err) => console.log(chalk`{red ${err}}\n\n`))
+      await Promise.all(serverErrorsWritten)
+      console.log('Please check the errors log for more details')
```
TimBeyer: Could we also log the name of the file so I can copy paste and `tail` it directly

src/bin/lib/write-errors-to-log.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const writeErrorsToLog = (errors: Error | Error[], filename: string): Promise<void> => {
+  if (!Array.isArray(errors)) {
+    errors = [errors]
+  }
+
+  let json = errors.map((err) => {
+    const errObj = {}
+    Object.getOwnPropertyNames(err).forEach((key) => errObj[key] = err[key])
+    return JSON.stringify(errObj, null, 2)
```
TimBeyer: This formats the JSON in a nicely readable way for people, but that also means there are plenty of newlines.
I think for the error log file we should render one object per line, especially since that's how newline delimited JSON works ;)

jfr3000: ah yes, i had human readability in mind but forgot about the newlines :woman_facepalming:

src/bin/lib/render-plan.ts
```diff
@@ -0,0 +1,24 @@
+import chalk from 'chalk'
+import { renderRuntimeErrorsAndWriteToFile, renderValidationErrors } from './plan-messages'
+import renderMessages from './plan-messages'
+import { RequestBatch } from '../../lib/offline-api/index'
+
+const renderPlan = async (batches: RequestBatch[], errorsFile: string): Promise<void> => {
```
madtrick: This function returns `undefined` (or `void`) instead of `Promise`, right?

src/bin/lib/plan-messages.ts
```diff
@@ -22,24 +23,37 @@ const renderPlan = (batches: RequestBatch[]) => {
...
     }
+  }
+}
 
+const renderValidationErrors = (batches: RequestBatch[]) => {
+  console.log(chalk`{bold.red The following migration has been planned but cannot be run because it contains errors}\n\n`)
+  renderPlan(batches)
+}
+
+const renderRuntimeErrorsAndWriteToFile = async (batches: RequestBatch[], filename): Promise<void> => {
```
madtrick: Missing type for `filename`

src/bin/lib/render-plan.ts
TimBeyer: It's an async function, so it returns a promise;)

src/bin/lib/render-plan.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+    console.log(chalk`🚨  {bold.red Migration unsuccessful}`)
+    process.exit(1)
+  }
+
+  console.log(chalk`{bold.green The following migration has been planned}\n`)
+  renderMessages(batches)
+}
+
+export default renderPlan
```
madtrick: I think this function is doing too many things at once:

- It'll render the plan
- (potentially) it'll write the errors to a file
- (potentially) it'll exit the process

What do you think if we pull the error handling code out of `renderPlan` so we endup with something like this:

(in `src/bin/cli.ts`)

```javascript
const errorsInPlan = batches.getErrors() // or something else

if (errorsInPlan) {
   await renderErrors(errorsInPlan) // this does the "if (hasValidationErrors || hasRuntimeErrors) {" logic

  process.exit(1)
}

await renderPlannedMigration(batches, errorsFile)

```

src/bin/lib/plan-messages.ts
madtrick: Related to some other comment I made, wdyt about extracting the `writeErrorsToLog(batch.runtimeErrors, filename)` logic to somewhere else? So the function does just one thing.

src/bin/lib/render-plan.ts
madtrick: You're right. Damm `asyn/await`

jfr3000:
To make handling of the parse result (which can have batches, or several types of errors, or both) easier, I have wrapped it in a class. Since we can now ask the `parseResult` about what types of errors occurred, we don't have to catch errors anymore.
Renderers are now purely responsible for rendering. File operations and terminating the process all happen in `cli.js`.

src/lib/migration-parser.ts
```diff
@@ -60,7 +94,12 @@ const createMigrationParser = function (fetcher): (migrationCreator: (migration:
 
     const ctsWithEntryInfo = await fetcher.checkContentTypesForDeletedCts(intentList, contentTypes)
 
-    validateChunks(intentList, ctsWithEntryInfo)
+    const payloadValidationErrors = validateChunks(intentList, ctsWithEntryInfo)
```
madtrick: nit (also not change on this PR), it feels a bit strange that on line 60 we do `intentList.validate()` and here we pass the `intentList` to this function for further validation.

test/integration/migration.spec.js
```diff
@@ -376,17 +377,10 @@ describe('the migration', function () {
...
-  it('throws an error when the script is invalid', co(function * () {
-    let result;
-
-    try {
-      result = yield migrationParser(invalidScript);
-    } catch (err) {
-      expect(err).to.be.an('error');
-      expect(err.message).to.include('Validation failed');
-    }
+  it('returns an error when the script is invalid', co(function * () {
```
madtrick: I think we should (if we've some time left) do a couple extra things:

- Rename this test to explicitly mention that it's only about `payloadValidationErrors`
- Add some tests for the other possible errors included in the `payloadResult`

jfr3000:
We agreed to address the above comments in a later PR

PR #53, opened 2017-11-23
Allow absolute script paths

zcei:


test/end-to-end/flags.spec.js
```diff
@@ -32,4 +34,28 @@ describe('contentful-migration CLI flags', function () {
...
+
+      cli()
+        .run(`--space-id ${SOURCE_TEST_SPACE} ${scriptPath}`)
+        .expect(assert.help.nonExistingMigrationScript(scriptPath))
+        .end(done);
+    });
+
+    it('resolves relative script path to cwd', function (done) {
+      const scriptPath = 'test/end-to-end/migration-scripts/non-existing-migration.js';
+      const resolvedPath = path.join(buildDir, scriptPath);
```
madtrick: 🙈  how is this path different to the one in line 43?

stupchiy: I also had the same question 😄

zcei: First test gets an "already resolved" (aka: absolute) path and expects it to not change.
This test uses a relative path and expects the CLI to resolve it to an absolute path.

madtrick: Ah, I see. We are testing the error response 👍

src/bin/cli.ts
```diff
@@ -20,7 +20,8 @@ const argv = yargs
   .usage('Parses and runs a migration script on a Contentful space.\n\nUsage: contentful-migration [args] <path-to-script-file>\n\nScript: path to a migration script.')
   .demandCommand(1, 'Please provide the file containing the migration script.')
   .check((args) => {
-    const filePath = path.join(process.cwd(), args._[0])
+    const rawScriptPath = args._[0]
+    const filePath = path.isAbsolute(rawScriptPath) ? rawScriptPath : path.join(process.cwd(), rawScriptPath)
```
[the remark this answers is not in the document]
zcei: https://www.pugski.com/media/homer-simpson-doh.499/full

PR #55, opened 2017-11-23
handle undefined returned values from transform and derive

jfr3000:
When `undefined` is returned from the transform or derive function, we don't produce a value for that locale on the transformed/ derived entry. If the function returns `undefined` for all locales of an entry, in the case of transform, it doesn't get changed, and in the case of derive, no target entry is created and the source entry is left untouched. (This is documented already in the README)

Verbose comments on this one, I can reduce those if needed.

### DIFF
--- a/src/lib/action/entry-derive.ts
+++ b/src/lib/action/entry-derive.ts
@@ -27,39 +27,67 @@ class EntryDeriveAction extends APIAction {
   async applyTo (api: OfflineAPI) {
     const entries: Entry[] = await api.getEntriesForContentType(this.contentTypeId)
     const locales: string[] = await api.getLocalesForSpace()
+
     for (const entry of entries) {
       const inputs = _.pick(entry.fields, this.fromFields)
       const newEntryId = await this.identityKey(inputs)
       const hasEntry = await api.hasEntry(newEntryId)
+
+      let skipEntry = true
+      let fieldsForTargetEntry = {}
+
+      for (const locale of locales) {
+        let outputsForCurrentLocale
+        try {
+          outputsForCurrentLocale = await this.deriveEntryForLocale(inputs, locale)
+        } catch (err) {
+          await api.recordRuntimeError(err)
+          continue
+        }
+
+        if (outputsForCurrentLocale === undefined) {
+          continue
+        }
+
+        skipEntry = false
+
+        // we collect all the values for the target entry before writing it to the
+        // offline API because we don't yet know if the entry might be skipped
+        // TODO: verify that the derivedFields actually get written to
+        // and to no other field
+        for (const [fieldId, localizedValue] of _.entries(outputsForCurrentLocale)) {
+          if (!fieldsForTargetEntry[fieldId]) {
+            fieldsForTargetEntry[fieldId] = {}
+          }
+          fieldsForTargetEntry[fieldId][locale] = localizedValue
+        }
+      }
+
+      // if derive returned undefined for all locales of this entry, there are no changes
+      // to be made, neither on the source nor the target entry, so we move on to the next
+      if (skipEntry) {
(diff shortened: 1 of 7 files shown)


src/lib/action/entry-derive.ts
```diff
@@ -27,40 +27,68 @@ class EntryDeriveAction extends APIAction {
...
+      for (const locale of locales) {
+        let outputsForCurrentLocale
+        try {
+          outputsForCurrentLocale = await this.deriveEntryForLocale(inputs, locale)
+        } catch (err) {
+          await api.recordRuntimeError(err)
+          continue
+        }
+
+        if (typeof outputsForCurrentLocale === 'undefined') {
```
madtrick: why this and not just `outputsForCurrentLocale === undefined`?

src/lib/action/entry-derive.ts
```diff
@@ -27,40 +27,68 @@ class EntryDeriveAction extends APIAction {
...
+          continue
+        }
+
+        skipEntry = false
+
+        // we collect all the values for the target entry before writing it to the
+        // offline API because we don't yet know if the entry might be skipped
+        // TODO: verify that the derivedFields actually get written to
+        // and to no other field
+        Object.keys(outputsForCurrentLocale).forEach((fieldId) => {
```
zcei: Minor styling:
```js
for ([fieldId, localizedValue] of entries(outputsForCurrentLocale))
```
makes it a bit clearer to me on why we need all `Object.keys` in each iteration.

src/lib/action/entry-derive.ts
```diff
@@ -27,40 +27,68 @@ class EntryDeriveAction extends APIAction {
...
-          }
-          // TODO: verify that the derivedFields actually get written to
-          // and to no other field
-          Object.keys(outputsForCurrentLocale).forEach((fieldId) => {
-            if (!targetEntry.fields[fieldId]) {
-              targetEntry.setField(fieldId, {})
-            }
-            targetEntry.setFieldForLocale(fieldId, locale, outputsForCurrentLocale[fieldId])
+
+          Object.keys(fieldsForTargetEntry[fieldId]).forEach((locale) => {
```
zcei: Same here - having `[fieldId, localizedField]` in line 82, and here `[locale, localizedValue]` makes the nested iteration a bit easier to digest.

src/lib/interfaces/request.ts
```diff
@@ -2,7 +2,10 @@ interface HttpRequest {
   method: 'PUT' | 'POST' | 'DELETE' | 'GET'
   url: string
   headers?: object
-  data?: object
+  data?: {
+    sys?: object,
```
zcei: For which kind of requests we need to send `sys`? 🤔

src/lib/action/entry-derive.ts
zcei: Because `undefined` could be overwritten (treated as variable), so you'd need to `outputsForCurrentLocale === void 0` here :trollface:

jfr3000: No good reason :innocent:

src/lib/interfaces/request.ts
jfr3000: That is not supposed to be there, thanks!

src/lib/action/entry-derive.ts
jfr3000: That reads much better indeed

src/lib/action/entry-derive.ts
madtrick: 🤷‍♂️ 

From what I understand TS transpiles by default to JS in strict mode so I would simplify and use the `outputsForCurrentLocale === undefined`

PR #58, opened 2018-01-12
Editor interfaces

thebiglabasky:
Proposed basic support for editor interfaces.

### DIFF
--- a/src/lib/fetcher.ts
+++ b/src/lib/fetcher.ts
@@ -1,18 +1,23 @@
 import IntentList from './intent-list'
-import APIContentType from '../lib/interfaces/content-type'
+import APIContentType, { APIEditorInterfaces } from '../lib/interfaces/content-type'
 import { ContentType } from '../lib/entities/content-type'
 import * as _ from 'lodash'
 import * as Bluebird from 'bluebird'
+import APIFetcher from './interfaces/api-fetcher'
 
-export default class Fetcher {
+export default class Fetcher implements APIFetcher {
   private makeRequest: any
 
   constructor (makeRequest) {
     this.makeRequest = makeRequest
   }
 
-  async getEntriesInIntents (intentList: IntentList) {
-    const ids: string[] = _.uniq(intentList.getIntents().map((intent) => intent.getContentTypeId()))
+  async getEntriesInIntents (intentList: IntentList): Promise<Array<Object>> {
+    const ids: string[] = _.uniq(
+      intentList.getIntents()
+      .filter((intent) => intent.isContentTransform() || intent.isEntryDerive())
+      .map((intent) => intent.getContentTypeId())
+    )
 
     if (ids.length === 0) {
       return []
@@ -27,10 +32,13 @@ export default class Fetcher {
   }
 
   async getContentTypesInChunks (intentList: IntentList): Promise<APIContentType[]> {
-    const ids: string[] = _.uniq(intentList.getIntents().reduce((ids, intent) => {
-      const intentIds = intent.getRelatedContentTypeIds()
-      return ids.concat(intentIds)
-    }, []))
+    const ids: string[] = _.uniq(intentList.getIntents()
+      .filter((intent) => !intent.isEditorInterfaceUpdate())
+      .reduce((ids, intent) => {
+        const intentIds = intent.getRelatedContentTypeIds()
+        return ids.concat(intentIds)
+      }, [])
+    )
 
     if (ids.length === 0) {
       return []
@@ -45,7 +53,33 @@ export default class Fetcher {
     return contentTypes
   }
 
-  async getLocalesForSpace () {
+  async getEditorInterfacesInIntents (intentList: IntentList): Promise<Map<string, APIEditorInterfaces>> {
+    const contentTypeIds: string[] = _.uniq(
+      intentList.getIntents()
+        .filter((intent) => intent.isEditorInterfaceUpdate())
+        .reduce((ids, intent) => {
+          const intentIds = intent.getRelatedContentTypeIds()
+          return ids.concat(intentIds)
+        }, [])
+    )
+
+    let editorInterfaces = new Map<string, APIEditorInterfaces>()
+
+    if (contentTypeIds.length === 0) {
+      return editorInterfaces
+    }
+
+    for (let id of contentTypeIds) {
+      const response = await this.makeRequest({
+        method: 'GET',
+        url: `/content_types/${id}/editor_interface`
+      })
+      editorInterfaces.set(id, response)
+    }
+    return editorInterfaces
+  }
+
+  async getLocalesForSpace (): Promise<Array<string>> {
     const response = await this.makeRequest({
       method: 'GET',
       url: `/locales`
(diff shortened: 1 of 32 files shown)


src/lib/fetcher.ts
```diff
@@ -1,18 +1,23 @@
 import IntentList from './intent-list'
-import APIContentType from '../lib/interfaces/content-type'
+import APIContentType, { APIEditorInterfaces } from '../lib/interfaces/content-type'
```
zcei: `import { APIContentType, APIEditorInterfaces } from ...`

src/lib/fetcher.ts
```diff
@@ -1,18 +1,23 @@
...
   constructor (makeRequest) {
     this.makeRequest = makeRequest
   }
 
-  async getEntriesInIntents (intentList: IntentList) {
-    const ids: string[] = _.uniq(intentList.getIntents().map((intent) => intent.getContentTypeId()))
+  async getEntriesInIntents (intentList: IntentList): Promise<Array<Object>> {
+    const ids: string[] = _.uniq(
+      intentList.getIntents()
+      .filter((intent) => intent.isContentTransform() || intent.isEntryDerive())
```
zcei: Nice catch 👏

src/lib/fetcher.ts
```diff
@@ -27,10 +32,13 @@ export default class Fetcher {
   }
 
   async getContentTypesInChunks (intentList: IntentList): Promise<APIContentType[]> {
-    const ids: string[] = _.uniq(intentList.getIntents().reduce((ids, intent) => {
-      const intentIds = intent.getRelatedContentTypeIds()
-      return ids.concat(intentIds)
-    }, []))
+    const ids: string[] = _.uniq(intentList.getIntents()
+      .filter((intent) => !intent.isEditorInterfaceUpdate())
```
zcei: Per se this is correct, though I think it makes it harder to understand, as it implies EditInterface updates wouldn't need to know anything about ContentTypes, which isn't correct, right?
Given that the EditorInterface update happens on a ContentType, the ContentType will be fetched either way, so I would rather let the `_.uniq` sort out that duplication.

src/lib/fetcher.ts
```diff
@@ -45,7 +53,47 @@ export default class Fetcher {
...
+
+    for (let id of contentTypeIds) {
+      try {
+        const response = await this.makeRequest({
+          method: 'GET',
+          url: `/content_types/${id}/editor_interface`
+        })
+        editorInterfaces.set(id, response)
+      } catch (error) {
+        if (error.name === 'NotFound') {
```
zcei: We should think about exposing the status code, that might look a bit cleaner than checking against our own API error codes.

src/lib/fetcher.ts
```diff
@@ -45,7 +53,47 @@ export default class Fetcher {
...
+
+    let editorInterfaces = new Map<string, APIEditorInterfaces>()
+
+    if (contentTypeIds.length === 0) {
+      return editorInterfaces
+    }
+
+    for (let id of contentTypeIds) {
+      try {
+        const response = await this.makeRequest({
```
zcei: For readability we should move this to it's own method, as this is the only place where we currently allow resources to not exist, which makes the code a bit bloated.

But correct approach 👍

src/lib/fetcher.ts
```diff
@@ -45,7 +53,47 @@ export default class Fetcher {
...
+          })
+        } else {
+          throw error
+        }
+      }
+    }
+    return editorInterfaces
+  }
+
+  async getLocalesForSpace (): Promise<Array<string>> {
```
zcei: We're trying to get rid of the `Array` generic, so please:
`Promise<string[]>`

src/lib/intent/base-intent.ts
```diff
@@ -115,6 +121,20 @@ export default abstract class Intent implements IntentInterface {
...
+   * @returns The API action to perform on save. Defaults to saving the Content Type.
+   */
+  getSaveAction (): APIAction {
+    return new ContentTypeSaveAction(this.getContentTypeId())
+  }
+
+  /**
+   * @returns The API action to perform on publish. Defaults to publishing the Content Type.
+   */
+  getPublishAction (): APIAction {
```
zcei: This block doesn't seem to be used at all?
Also you can just import the `ContentTypeSaveAction` / `ContentTypePublishAction` in your intent and add it as an action. See `src/lib/intent/field-delete.ts` for an example.

src/lib/intent/composed-intent.ts
```diff
@@ -36,6 +37,15 @@ export default class ComposedIntent implements Intent {
   }
 
   // TODO: adjust interface so we don't have to implement all this
+  getSaveAction (): APIAction {
+    return null
+  }
+  getPublishAction (): APIAction {
```
zcei: This just mirrors the interface from above and therefore isn't used as well.

src/lib/intent/editorinterface-update.ts
```diff
@@ -0,0 +1,44 @@
...
+    return true
+  }
+  shouldSave (): boolean {
+    return false
+  }
+  shouldPublish (): boolean {
+    return false
+  }
+  toActions () {
+    // TODO: check for optimizing API calls and saving CT by CT with merged editor interface definition
```
zcei: Is this possible? It is its on API call, never seen it merged into a ContentType update

src/lib/offline-api/index.ts
```diff
@@ -1,7 +1,7 @@
 import { omit, compact } from 'lodash'
 import FieldDeletionValidator from './validator/field-deletion'
 import { ContentTypePayloadValidator } from './validator/content-type'
-import ContentType from '../entities/content-type'
+import ContentType, { EditorInterfaces } from '../entities/content-type'
```
zcei: `import { ContentType, EditorInterfaces } from ...`

src/lib/intent/composed-intent.ts
thebiglabasky: Ah yes, remains from an attempt to make the save content type and publish content type actions generic.
I'll get rid of these

src/lib/intent/editorinterface-update.ts
thebiglabasky: Nah, I wanted ideally to merge editor interface changes of the same content type in a single API call, but that doesn't seem to be easy to do.
Maybe as a follow up!

src/lib/fetcher.ts
thebiglabasky: In theory, you don't need to fetch the content type if the only thing you do is updating the editor interface though.
I'd suggest we keep, and add a comment?

src/lib/fetcher.ts
thebiglabasky: I'll add a TODO

src/lib/fetcher.ts
zcei: Ah you're right, the `editContentType` doesn't dispatch an intent.

thebiglabasky:
@zcei Could you PTAL? Thanks!

PR #61, opened 2018-01-18
Use stderr for warnings/errors

### LINKED ISSUE #60: Use stderr for errors
Currently all errors are sent to stdout (via `console.log`), all errors should go to stderr (via `console.error`)

hhsnopek:
closes #60


src/bin/lib/render-migration.ts
```diff
@@ -28,28 +28,28 @@ const renderPlan = (batches: RequestBatch[]) => {
     }
 
     for (const error of batch.validationErrors) {
-      console.log(chalk`{red.bold Error: ${error.message}}`)
+      console.error(chalk`{red.bold Error: ${error.message}}`)
     }
 
     if (batch.intent.shouldPublish()) {
-      console.log(chalk`\n{bold.underline Publish Content Type} {bold.yellow ${batch.intent.getContentTypeId()}}`)
+      console.error(chalk`\n{bold.underline Publish Content Type} {bold.yellow ${batch.intent.getContentTypeId()}}`)
```
TimBeyer: This isn't an error case. It's part of the normal plan output.

hhsnopek: Updated! I assumed that the `yellow` color inferred warning/error in this case

TimBeyer:
It seems like running integration tests does not work correctly from forks.  
I assume that's because of security reasons, which makes a lot of sense actually.
I guess we will have to commit the changes into a branch that's in this repo and proceed from there.

PR #65, opened 2018-02-20
Allow null values for flags in regexp validations

### LINKED ISSUE #64: [Bug] Migration fails when content model has existing regex validations without flags
I have a straightforward migration I'm trying to run to add a field to a content type.
```javascript
     const page = migration.editContentType('article');

    page.createField('stickyCTANavBar')
      .type('Link')
      .linkType('Entry')
      .name('Sticky CTA Nav Bar')
      .validations([{ linkContentType: ['ctaConfig'] }])
 };
```

The output when I run it:
```
Publish Content Type article
Update Content Type financialAdvisorLandingPage

  Create field stickyCTANavBar
    - type: "Link"
    - linkType: "Entry"
    - name: "Sticky CTA Nav Bar"
    - validations: [{"linkContentType":["ctaConfig"]},{"flags":""}]


Error: "flags" validation expected to be "string", but got "null"
```

I have tried removing the content type validation, but I still get this error. Parsing through the documentation, I see there is a flag parameter for regex validations, but I'm not doing any regex validations on this field. If there is something wrong with the content model's other fields, it is not noticeable from the Contentful UI, since the model seems to be valid and working as expected. Any ideas?

PepperTeasdale:
When a regexp validation is created through contentful ui, if no flag is
provided, it is set as `null`. After that, you are unable to update that
content model via migration-cli. This allows null values for that
validation property. Fixes #64 

I wasn't sure exactly where to put the test for this, but I think this was a reasonable place. Please advise if you prefer anything different.

### DIFF
--- a/src/lib/offline-api/validator/schema/field-validations-schema.ts
+++ b/src/lib/offline-api/validator/schema/field-validations-schema.ts
@@ -18,7 +18,7 @@ const rangeValidation = validation('range', range('number'))
 
 const regexp = validation('regexp', Joi.object({
   pattern: Joi.string(),
-  flags: Joi.string()
+  flags: Joi.string().allow(null).optional()
 }))
 
 const unique = validation('unique', Joi.boolean())
--- a/test/unit/lib/offline-api/validation/payload-validation-field-validations.spec.ts
+++ b/test/unit/lib/offline-api/validation/payload-validation-field-validations.spec.ts
@@ -131,5 +131,22 @@ describe('payload validation', function () {
         []
       ])
     })
+
+    it('allows null values for optional validation properties', async function () {
+      const errors = await validateBatches(function up (migration) {
+        const person = migration.createContentType('person')
+          .name('Person')
+          .description('A Person')
+
+        person.createField('fullName')
+          .name('Full Name')
+          .type('Symbol')
+          .validations([{ regexp: { pattern: '^[A-Za-z\s]+$', flags: null } }])
+      }, [])
+
+      expect(errors).to.eql([
+        []
+      ])
+    })
   })
 })


src/lib/offline-api/validator/schema/field-validations-schema.ts
```diff
@@ -18,7 +18,7 @@ const rangeValidation = validation('range', range('number'))
 
 const regexp = validation('regexp', Joi.object({
   pattern: Joi.string(),
-  flags: Joi.string()
+  flags: Joi.string().allow(null)
```
TimBeyer: I'm wondering if we should also add `.optional()`. 
It seems like the API also accepts `flags` to be left out.

PepperTeasdale: Happy to add that if desired. When I initially wrote the test, I simply left off the `flags` key from the validation object and the test passed, so it seems like it was effectively optional. Only after I explicitly set the value to `null` did the test fail. The test is a little contrived, since it's unlikely anyone would try to explicitly set a null value on `flags`, and the validation was failing against fields on the entry returned from the API.

PepperTeasdale: Also, travis is failing because
```
Error: Please define all necessary "CONTENTFUL_INTEGRATION_*" environment variables
```
I'm a little confused, since that seems unrelated to this PR. Please advise if I need to be doing something differently here.

TimBeyer: That's happening because the integration tests need access tokens defined, and for security reasons travis will not expose access tokens to builds triggered by forks of the main repo.  
Otherwise anyone could make a PR that just logs the environment variables and the secrets are leaked. 
That does make things very annoying for anyone proposing a PR though, so we'll eventually have to look into a better solution. For now, don't worry about it.

PepperTeasdale:
@TimBeyer 💇

PepperTeasdale:
@TimBeyer anything blocking a merge of this?

Soliah:
I've forked and manually applied this change and it works great. Be great to get this merged and a release cut.

thebiglabasky:
Hi,
Sorry for the delay. The team is currently massively hit by the flu and didn't have much room to go through this.
We'll give you an update as soon as possible.

Thanks for your understanding.

Hervé

jody-zeitler:
I've been keeping an eye on this one for a while. @PepperTeasdale can you merge master and get it back on the radar?

Khaledgarbaya:
Hi @PepperTeasdale Sorry for the delay I will get to this first thing tomorrow

PR #67, opened 2018-02-21
Fix import causing logger to crash

willm:
I was getting the following error when running this from my machine, I don't think there is a package called util. I think the intention was to use bluebird (which seemed to work).

```
contentful-migration --space-id $$$ migration.js                                                                                                                                                                                                                                                                                                    /usr/local/lib/node_modules/contentful-migration-cli/built/bin/lib/write-errors-to-log.js:5                                                                                                                                                                                                                                                                               const appendFile = util.promisify(fs.appendFile);                                                                                                                                                                                                                                                                                                                                                 ^                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           TypeError: util.promisify is not a function
at Object.<anonymous> (/usr/local/lib/node_modules/contentful-migration-cli/built/bin/lib/write-errors-to-log.js:5:25)
at Module._compile (module.js:571:32)
at Object.Module._extensions..js (module.js:580:10)
at Module.load (module.js:488:32)
at tryModuleLoad (module.js:447:12)
at Function.Module._load (module.js:439:3)
at Module.require (module.js:498:17)
at require (internal/module.js:20:19)
at Object.<anonymous> (/usr/local/lib/node_modules/contentful-migration-cli/built/bin/cli.js:15:31)
at Module._compile (module.js:571:32)
at Object.Module._extensions..js (module.js:580:10)
at Module.load (module.js:488:32)
at tryModuleLoad (module.js:447:12)
at Function.Module._load (module.js:439:3)                                                                                                                                                                            at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
```

### DIFF
--- a/src/bin/lib/write-errors-to-log.ts
+++ b/src/bin/lib/write-errors-to-log.ts
@@ -1,5 +1,5 @@
 import * as fs from 'fs'
-import * as util from 'util'
+import * as util from 'bluebird'
 
 const appendFile = util.promisify(fs.appendFile)
 


zcei:
Hey @willm,

the "package" called `util` is a Node internal one: https://nodejs.org/api/util.html

The `promisify` method was added in Node 8, which is the required version to run this project:
https://github.com/contentful/migration-cli/blob/4836f3ef68fcccda364227db316cb8f58c95db86/package.json#L120-L122

What version of Node did you use? Wondering, because the installation should have already caused a warning.

willm:
I managed to install it with `npm install -g contentful-migration-cli` with node version 7.10.0. according to [npm](https://docs.npmjs.com/files/package.json#enginestrict) if you want to enforce the version, you should use enginesStrict. However, using bluebird, it did work in node 7. So maybe it's worth considering that, since you're using it anyway?

zcei:
Even though it might be working right now, we don't make any promises it will still work with upcoming changes.

What's the reason you can't install the latest LTS (Node 8)? Node 7 actually reached its end of life in July '17, so there's no reason for us to further support it.

Also it seems like `engineStrict` has been removed in `npm@3`.

willm:
OK I will update.

PR #77, opened 2018-03-05
Allow null values for flags in regexp validations

Soliah:
When a regexp validation is created through contentful ui, if no flag is
provided, it is set as `null`. After that, you are unable to update that
content model via migration-cli. This allows null values for that
validation property.


Soliah:
Oops was trying to do this our fork. Ignore.

PR #83, opened 2018-04-06
Fetch all entries to transform

### LINKED ISSUE #82: [Bug] Only first 100 entries are transformed
Hello 👋 

I’m running into an issue when using the `transformEntries()` method on a large number of entries. Only the first 100 entries are transformed. I assume this is due to the paginated response from the management API with a default limit of 100.

@stefanjudis has pointed me to https://github.com/contentful/migration-cli/blob/master/src/lib/fetcher.ts#L26-L29, so I'll give it a shot and make a PR.

connor-baer:
Fixes #82.

### DIFF
--- a/src/lib/fetcher.ts
+++ b/src/lib/fetcher.ts
@@ -1,5 +1,6 @@
 import IntentList from './intent-list'
 import { APIContentType, APIEditorInterfaces } from '../lib/interfaces/content-type'
+import APIEntry from '../lib/interfaces/api-entry'
 import { ContentType } from '../lib/entities/content-type'
 import * as _ from 'lodash'
 import * as Bluebird from 'bluebird'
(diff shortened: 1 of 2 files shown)


src/lib/fetcher.ts
```diff
@@ -23,12 +24,24 @@ export default class Fetcher implements APIFetcher {
...
+    let skip: number = 0
+
+    while (true) {
+      const response = await this.makeRequest({
+        method: 'GET',
+        url: `/entries?sys.contentType.sys.id[in]=${ids.join(',')}&skip=${skip}`
+      })
+
+      entries = entries.concat(response.items)
+      skip += 100
```
TimBeyer: I'd prefer not to have a magic number here and use `response.items.length`.

connor-baer: Good point!

connor-baer:
@TimBeyer I'll add some tests after lunch.

Khaledgarbaya:
LGTM although at some point we will change the code to not have all entries in memory at once.
@zcei I think we can merge this.

Khaledgarbaya:
@connor-baer Thanks for the PR we'll release a `next` version soon

PR #88, opened 2018-04-18
Update unit tests to run on windows

Roblinde:


Roblinde:
@Khaledgarbaya could you have a look and make sure this doesn't miss any tests or breaks something for mac/linux users.

Thanks!

PR #97, opened 2018-05-09
Add Editor Interfaces Settings

### LINKED ISSUE #99: Enable bulk editing
It would be useful if bulk editing flag could be set as part of the content model definition. 

At present I can't see a way to do this.

thanks!

dlitvakb:
Fixes #99

### DIFF
--- a/src/lib/action/editorinterface-update.ts
+++ b/src/lib/action/editorinterface-update.ts
@@ -1,13 +1,14 @@
 import { EditorInterfaces } from '../entities/content-type'
+import { APIEditorInterfaceSettings } from '../interfaces/content-type'
 import { EntityAction } from './action'
 
 class UpdateEditorInterfaceAction extends EntityAction {
   protected contentTypeId: string
   protected fieldId: string
   private _widgetId: string
-  private _settings?: object
+  private _settings?: APIEditorInterfaceSettings
 
-  constructor (contentTypeId: string, fieldId: string, widgetId: string, settings?: object) {
+  constructor (contentTypeId: string, fieldId: string, widgetId: string, settings?: APIEditorInterfaceSettings) {
     super()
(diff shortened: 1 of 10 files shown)


src/lib/action/editorinterface-update.ts
```diff
@@ -5,12 +5,14 @@ class UpdateEditorInterfaceAction extends EntityAction {
   protected contentTypeId: string
   protected fieldId: string
   private _widgetId: string
+  private _settings?: object
```
Khaledgarbaya: shouldn't this be `APIEditorInterfaceSettings` instead of object

src/lib/action/editorinterface-update.ts
```diff
@@ -5,12 +5,14 @@ class UpdateEditorInterfaceAction extends EntityAction {
...
 
-  constructor (contentTypeId: string, fieldId: string, widgetId: string) {
+  constructor (contentTypeId: string, fieldId: string, widgetId: string, settings?: object) {
```
Khaledgarbaya: same here `APIEditorInterfaceSettings `

src/lib/entities/content-type.ts
```diff
@@ -98,20 +95,39 @@ class EditorInterfaces {
...
-    let fieldId: string
-    for ([fieldId, widgetId] of this._controls.entries()) {
+    let result: APIEditorInterfaceControl[] = []
+    forEach(this._controls, (c) => {
       result.push({
-        fieldId,
-        widgetId
+        fieldId: c.fieldId,
+        widgetId: c.widgetId,
+        settings: c.settings
```
thebiglabasky: What if there's no `c.settings`? shouldn't the payload remove the property entirely instead of sending `settings: undefined`?

PR #100, opened 2018-05-12
Improvements to editor interface migrations

connor-baer:
## Changes 

_When changing the editor interface of a field, I'd like to see those changes listed in the intents before approving the migration._

- Include field settings in editor interface intent
- Fix true/false label types
- Extend documentation for `changeEditorInterface`

Screenshot of the added details when updating an editor interface

> I'm not sure if this is the proper way of doing it — one thing that bothers me is that there's no spacing between the editor interface intent and the next intent. I couldn't figure out how to add it.
> Happy to be pointed in the right direction :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,21 +267,31 @@ Edits the field of provided `id`.
 Shorthand method to omit a field, publish its content type, and then delete the field.
 This implies that associated content for the field will be lost.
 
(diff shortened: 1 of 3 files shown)


Khaledgarbaya:
Thanks @connor-baer

PR #109, opened 2018-06-21
Update Readme

stefanjudis:
As we don't recommend using this module as CLI anymore global installation instructions may be removed. :)


Khaledgarbaya:
Thanks @stefanjudis

PR #116, opened 2018-07-16
Update dependencies

Khaledgarbaya:
This PR update vulnerable dependencies exept `string` inside of nixt because there is no fix released yet and also based on [Snyk](https://snyk.io/vuln/npm:string:20170907) report the affected code is inside `underscore` and `unescapeHTML` which are not used inside of nixt so we should be fine


Khaledgarbaya:
Hey @axe312ger, CI is 💚 can you take a look

Issue #111, opened 2018-06-26
Catching errors or aborted runs when using migration CLI as a library


benjarwar:
## Expected Behavior

When use the CLI tool as a library and running a migration, I need to catch both:

1. When a migration fails
2. When the user hits `n` when prompted to proceed with the migration

## Actual Behavior

The function currently does not throw an error or return a Promise that would allow me to catch these cases. Therefore, when I run a migration via my Node scripts, subsequent code that should only run on successful migration is running every time.

## Possible Solution

The returns should for each case could perhaps be improved:

- Success [currently returns the results of Listr.run()](https://github.com/contentful/contentful-migration/blob/master/src/bin/cli.ts#L169) which in my testing is an empty `{}` Object. This object could be enhanced to provide a mechanism for checking results.
- The [error catch](https://github.com/contentful/contentful-migration/blob/master/src/bin/cli.ts#L175) doesn't return anything.
- The [migration aborted](https://github.com/contentful/contentful-migration/blob/master/src/bin/cli.ts#L178) condition doesn't return anything.

## Steps to Reproduce

Stripped down code:

```
const cfRunMigration = require("contentful-migration/built/bin/cli").default;

const runMigration = async () => {
  try {
    const options = {
      // redacted
    };

    await cfRunMigration(options);
  } catch (e) {
    // doesn't seem to fire on errors, because no error is thrown by cfRunMigration
    console.log(e);
  }

  // Always runs, even on error or when migration is aborted
  thisShouldNotAlwaysRun();
}
```

I can sort of solve for the aborted attempts by checking if the output of `cfRunMigration` is undefined; but that feels dirty:

```
const cfRunMigration = require("contentful-migration/built/bin/cli").default;

const runMigration = async () => {
  let migrationResult;

  try {
    const options = {
      // redacted
    };

    migrationResult = await cfRunMigration(options);
  } catch (e) {
    // doesn't seem to fire on errors, because no error is thrown by cfRunMigration
    console.log(e);
  }

  // Dirty check
  if (typeof migrationResult !== 'undefined') {
    thisShouldNotAlwaysRun();
  }
}
```

## Context

[...]

PR #118, opened 2018-07-17
feat: Add support for throwing instead of process.exit(1) on errors

charlespwd:
## Summary

When using the package as a library, it's a good idea to throw an error
instead of exiting with code 1. This way the user can recover from the
mistake somehow (e.g. deleting a throwaway environment on during continuous
integration).

This seems like it solves #111 too. 

## Description

- When using the `runMigration` export, make the function throw instead of process.exit(1).
- When using the `default` export, proceed as before by running process.exit(1) on errors. 

## Motivation and Context

I'm making continuous integration scripts that creates and deletes throwaway branches on CircleCI to test that the migration scripts are valid before they are merged. When the migration fails, I would like it to delete the newly created branch. This is currently impossible because this library will `process.exit(1)` for me. 

Actual run-migration code:
```js
async function main() {
  let statusCode = 0;
  try {
    await createEnvironment();
    await runMigrations();
  } catch (e) {
    console.error(e);
    statusCode = 1;
  } finally {
    await deleteEnvironment();
  }

  process.exit(statusCode);
}
```

Without this PR, the code above is unable to execute the `finally` block after errors.  

I'm open to feedback / suggestions on how to improve this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ npm install contentful-migration
 ### Usage as a library
 
 ```javascript
-const runMigration = require('contentful-migration/built/bin/cli').default
+const runMigration = require('contentful-migration/built/bin/cli').runMigration
 const options = {
   filePath: '<migration-file-path>',
   spaceId: '<space-id>',
@@ -446,7 +446,7 @@ You can check out the [examples](/examples) to learn more about the migrations D
 Each example file is prefixed with a sequence number, specifying the order in which you're supposed to run the migrations, as follows:
 
 ```javascript
-const runMigration = require('contentful-migration/built/bin/cli').default
+const runMigration = require('contentful-migration/built/bin/cli').runMigration
 
 const options = {
   spaceId: '<space-id>',
--- a/examples/usage-as-lib.js
+++ b/examples/usage-as-lib.js
@@ -1,0 +1,24 @@
+const { runMigration } = require('../built/bin/cli');
+
+async function main (filePath) {
+  let statusCode = 0;
+
+  try {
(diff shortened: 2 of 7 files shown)


charlespwd:
I don't know how to configure the CONTENTFUL_INTEGRATION_* on travis. But the tests pass for me!
https://pastebin.com/raw/4sc5bWAC

Khaledgarbaya:
Hi @charlespwd,
Thank you for the PR, we will do the test locally and come back to you. I am working on improving the integration test situation using fixtures which will solve the testing problem soon.

charlespwd:
Hi @Khaledgarbaya, @axe312ger. Any updates on this? We'd like to use this here at ALDO group. Thanks.

Khaledgarbaya:
Hi @charlespwd I will check this this first thing tomorrow.
Thanks

Khaledgarbaya:
Hi @charlespwd,
First look the PR is looking good but we would like to have some tests for this behaviour. Especially some integration test. You can check the existing tests and copy some code from there to get a boilerplate ready.
Let me know if I can help
Khaled

charlespwd:
Hey @Khaledgarbaya, the e2e tests fail for me on master. Is that normal?

charlespwd:
Well I didn't want to close it.

charlespwd:
OK, I got an e2e test in. Let me know if you need anything else :) 

@Khaledgarbaya @axe312ger

Khaledgarbaya:
available in `v0.11.1`

Issue #117, opened 2018-07-17
Token/SpaceId not exists even if they are


stefanzmindit:
## Expected Behavior

Run a contentful migration
## Actual Behavior

Not able to run contentful migrations (most basic one*), getting this error 
_The provided space does not exist or you do not have access_. Removing contentful-migration@**0.10.0** and switching back to **0.9.0** works as expected using the same spaceId and accessToken

*migration example:
```
module.exports = function(migration) {
    migration.createContentType('test', {
        'name': 'Test',
        'description': 'Test'
    });
}
```

## Environment

-   **Node Version**: v10.0.0
-   **Package Manager Version**: 5.6.0
-   **Operating System**: Darwin Mindits-MacBook-Pro.local 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64 x86_64
-   **Package Version**: 0.10.0

Khaledgarbaya:
Hey @stefanzmindit are you using the tool as a library or as the command line. Can you provide the usage snippet

stefanzmindit:
Hey,

I use it as library through a node script:

```
const runMigration = require('contentful-migration/built/bin/cli').default
const pathJoin = require('path').join;

function migrate() {
    runMigration({
        spaceId: 'SPACEID',
        accessToken: 'TOKEN',
        filePath: pathJoin(__dirname, 'files/migrate_1.js')
    })
        .then(() => console.log('Migration Done!'))
        .catch((e) => console.error)        
}

migrate();
```

and in files/migrate_1.js

```
module.exports = function(migration) {
    migration.createContentType('test', {
        'name': 'Test',
        'description': 'Test'
    });
}
```

package.json looks like

```
{
  "name": "migration",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "migrate": "node do.js"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "contentful": "^6.1.1",
    "contentful-migration": "^0.9.0"
  }
}
```

stefanzmindit:
Running the same script ```node run migrate``` with 0.10.0 gives me the error described, but replacing it with 0.9.0 (without doing anything else than npm uninstall contentful-migration and npm install contentful-migration@0.9.0) works fine

stefanjudis:
[...]

PR #119, opened 2018-07-18
fix: default env to master

Khaledgarbaya:
This PR defaults the environment to master

Closes https://github.com/contentful/contentful-migration/issues/117

### DIFF
--- a/src/bin/lib/render-migration.ts
+++ b/src/bin/lib/render-migration.ts
@@ -18,7 +18,8 @@ const renderBatch = function (batch: RequestBatch) {
   console.log(message.join('\n'))
 }
 
-const renderPlan = (batches: RequestBatch[]) => {
+const renderPlan = (batches: RequestBatch[], envrionment: string) => {
+  console.log(chalk`{bold.underline Environment}: {bold.yellow ${envrionment}}\n`)
   console.log(chalk`{bold.green The following migration has been planned}\n`)
   for (const batch of batches) {
     renderBatch(batch)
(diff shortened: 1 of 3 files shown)


connor-baer:
I wonder whether it would be better to force the user to provide an environment id. I can see myself accidentally running a migration on master with this change, simply because I forgot to provide the environment id.

Khaledgarbaya:
Hey @connor-baer some users might not have access to the environment feature and all of our other tools are defaulting to master for now, This might change when we reach the v1.0.0

zcei:
To adress @connor-baer's doubt, I'd really like to have a `Environment: <the-specified-environment>`  in the plan output. As the called URLs are relative to the environment you don't see what you're executing against during runtime.

Khaledgarbaya:
@connor-baer @zcei we now have the current environment displayed in the plan
screen shot 2018-07-26 at 14 11 47

src/bin/lib/render-migration.ts
```diff
@@ -19,8 +19,8 @@ const renderBatch = function (batch: RequestBatch) {
 }
 
 const renderPlan = (batches: RequestBatch[], envrionment: string) => {
-  console.log(chalk`{bold.underline Environment}: {bold.yellow ${envrionment}}\n`)
   console.log(chalk`{bold.green The following migration has been planned}\n`)
+  console.log(chalk`{bold.underline Environment}: {bold.yellow ${envrionment}}\n`)
```
connor-baer: Looks like you have a typo here: `${envrionment}` 😉

zcei: Haha, but thanks to auto-complete it's consistently wrong 😂

PR #123, opened 2018-07-19
Test: add recording

Khaledgarbaya:
This PR adds recording to the integration and en2en tests so users can run them on their machine without the need to hit the API endpoint


package.json
```diff
@@ -112,7 +113,8 @@
     "tslint": "^5.7.0",
     "tslint-config-standard": "^6.0.1",
     "typescript": "^2.5.3",
-    "uuid": "^3.2.1"
+    "uuid": "^3.2.1",
```
axe312ger: not sure but didn't you remove all the uuid usage?

Khaledgarbaya:
@axe312ger Can you 👀

test/record.js
```diff
@@ -0,0 +1,62 @@
+const nock = require('nock');
+const path = require('path');
+const fs = require('fs');
+const zlib = require('zlib');
+
+module.exports = function (name, options) {
+  // options tell us where to store our fixtures
+  options = options || {};
+  const fixtureFolder = options.fixtureFoldr || 'fixtures';
```
axe312ger: typo in options config `fixtureFoldr `

test/record.js
```diff
@@ -0,0 +1,62 @@
...
+  const fp = path.resolve(__dirname, path.join(fixtureFolder, name + '.js'));
+  // `hasFixture` indicates whether the test has fixtures we should read,
+  // or doesn't, so we should record and save them.
+  // the environment variable `NOCK_RECORD` can be used to force a new recording.
+  let hasFixture = process.env.NOCK_RECORD === 1;
```
axe312ger: Based on your comment I'd expect it to detect if a fixture is present or not. But you just hardcoded it based on the environment variable. When I look at the npm scripts, it is always 0, so hasFixture would always be false, creating new fixtures every run. How does this work? 🤔

test/record.js
```diff
@@ -0,0 +1,62 @@
...
+      } else {
+        hasFixture = false;
+        nock.recorder.rec({
+          dont_print: true
+        });
+      }
+    },
+    // saves our recording if fixtures didn't already exist
+    after: function (done) {
+      if (!hasFixture) {
```
axe312ger: Shouldnt you check this based if you have a recording available or not? Plus only record when you need to? 🤔

test/integration/migration.spec.js
```diff
@@ -18,9 +18,15 @@ const changeEditorInterfaceWithExistingContentTypeAddingHelpText = require('../.
...
 
-const ENVIRONMENT_ID = uuid.v4();
+const record = require('../record');
+
+const recorder = record('contentful-migration-integration');
+
+before(recorder.before);
+after(recorder.after);
+
+const ENVIRONMENT_ID = 'env-integration';
```
axe312ger: 👍 on how easy it is to integrate <3

test/record.js
Khaledgarbaya: Sometimes we want to refresh the fixtures if for example, we add a new endpoint so we change `NOCK_RECORD` to `1`

test/record.js
Khaledgarbaya: Same comment as above

axe312ger:
😻

PR #132, opened 2018-09-07
fix(field-validations): range validations allow null require min/max

### LINKED ISSUE #106: Cannot create validation with assetImageDimension values that are null.
Hello, I have an asset as a field in one of my content models, when I created the validations for it using the website it worked fine and this is the JSON representation it creates: 

```
{
  "assetImageDimensions": {
    "width": {
      "min": 1100,
      "max": null
    },
    "height": {
      "min": null,
      "max": null
    }
  }
}
```

However, if I try to create a migration using this package and I use null as a max or min value then the cli command errors and doesn't allow me to run the migration:

```
Error: "max" validation expected to be "number", but got "null"
Error: "min" validation expected to be "number", but got "null"
Error: "max" validation expected to be "number", but got "null"
```

If I simply omit it the null values like this:

```
{
  assetImageDimensions: {
    width: {
      min: 1100
    }
  }
}
```

Then the migration command runs successfully, but when I use the website I cannot edit the field, if I try I get an error "The application has encountered a problem and needs to reload".  I am speculating that this is because of the missing height/width or max/min values, the website needs them to be `null`, however I don't think it's entirely a problem with the website because clearly `null` values are allowed and the cli command erroneously prevents them from being specified.

phoebeschmidt:
fixes #106

## Summary

Update field validation schema for range validations. Now both `min` and `max` are required, and both allow null.


test/end-to-end/validations.spec.js
```diff
@@ -69,7 +66,7 @@ describe('apply validations migration examples', function () {
       .on(/\? Do you want to apply the migration \(Y\/n\)/).respond('y\n')
       .expect(assert.plans.contentType.create('dieatary-food', { name: 'Dieatary Food', description: 'Food with up to 500 calories' }))
       .expect(assert.plans.field.create('name', { type: 'Symbol', name: 'name of the food', validations: [{ unique: true }] }))
-      .expect(assert.plans.field.create('calories', { type: 'Number', name: 'amount of calories the food contains', validations: [{ range: { 'max': 500 } }] }))
+      .expect(assert.plans.field.create('calories', { type: 'Link', linkType: 'Asset', name: 'amount of calories the food contains', validations: [{ assetImageDimensions: { width: { min: 1199, max: null }, height: { min: null, max: null } } }] }))
```
Khaledgarbaya: It would be great if we have another test for the wrong case, for example, when one of the fields is omitted `{min: 1232}`

PR #133, opened 2018-09-10
Fix vulnrerable dependency

Khaledgarbaya:
## Summary

[string.js](https://github.com/jprichardson/string.js/issues/212) which is a dependency of [nixt](https://github.com/vesln/nixt/pull/18) is vulnerable.
This PR fixes that.
Even though the vulnerable code is not used it's good to remove it.


Khaledgarbaya:
@phoebeschmidt the recording is only used with integration test and not with e2e that's why

Issue #113, opened 2018-06-29
Allow transformEntries style migrations between different Content Models


Shy:
## Context

@benjarwar can jump in here with more context, but we're looking for the ability to allow transformEntries from the migration CLI but between different content models. 

For example, Content Model A has Author. We create a new Content Model B. Content Model B needs Author but split into first name, last name.

benjarwar:
@Shy that's pretty much the idea. Even directly porting fields from one entry type to a new entry type. In our current use case, we're looking to change the entry ID/name (and make some adjustments). Being able to port the content simultaneously would be a big time saver.

TimBeyer:
Is [deriveLinkedEntries](https://github.com/contentful/contentful-migration#derivelinkedentriesconfig) what you are looking for, or do you want the same functionality but without the linking part?

benjarwar:
@TimBeyer `deriveLinkedEntries` seems more specific to creating new entries out of existing field data. My use case is a little different. I actually need to port/convert an associated reference to a new entry type.

Given an entry type of `Parent`, which contains an `associatedChild` field which can be a reference to `Child1` entries. I've now created a new entry type, `Child2`, which should be used in place of `Child1`.

I would like to now:
- Iterate over all `Parent` entries
- If it contains a `Child1` reference in `associatedChild`, create a new `Child2` entry, copying over all data from the `Child1` entry (some fields will also be transformed in this step)
- Delete the original `Child1` entry
- Later, deprecate `Child1`, removing the entry type entirely from the system

benjarwar:
Researching further, I'm starting to think the solution lies in using the [Content Management API](https://www.contentful.com/developers/docs/references/content-management-api/) rather than trying to fold this into my migration scripts...

aKzenT:
While it's possible to do this all using the content management API, I would also appreciate a version of this directly in the migration CLI.

We often have the same use cases that @benjarwar mentions. We deprecate one content type in favour of another one or splitting up a content type into several ones.

Another use case would be renaming of content types. It's a real pain that it's impossible to rename contentTypes (sys.id not display name). We have a lot of content types and sometimes the name doesn't match the use cases or our naming conventions anymore. With migrations we could at least create a new content type with the same fields and migrate all entries over.

Thanks!

aKzenT:
Hey,

thanks for the response. I would image something like this:

```
  migration.migrateEntries({
    contentType: 'sourceType',
    targetContentType: 'targetType',
[...]

PR #143, opened 2018-10-19
New Action to transform entries to another content type

aKzenT:
## Summary
This PR extends the available migrations with a new "transformEntriesToType"-transform that allows migrations from one content type to another as discussed in https://github.com/contentful/contentful-migration/issues/113

## Description
A new action was added to support the transformation. Various enhancements to the OfflineAPI as well as the Fetcher were needed to implement the features.

## Motivation and Context
See https://github.com/contentful/contentful-migration/issues/113

## Todos

-   [x] Basic functionality working
-   [x] Option to remove the source item after transformation
-   [x] Option to publish/not publish or keep the state of the original entry
-   [x] Option to change all references to the original entry so that they point to the new entry instead
-   [x] Additional tests


Khaledgarbaya:
Hey @aKzenT,
😱 this looks amazing, thanks for the contribution.
Let me know when this is ready for review and we will go through it.

Best,
Khaled

aKzenT:
We have no additional changes planned for the moment, so it would be great if you can start the review so we can get this into the product.

aKzenT:
Hey. Do you have any ETA for this to be merged? We have spent a lot of time and effort to develop this and would love to see it integrated into the tooling soon, so we can take advantage of it. Thanks!

jesustalaveragomez:
Hi, are you planning to release this feature soon?

phoebeschmidt:
Hey @aKzenT and @jesustalaveragomez, sorry for the delay. We know the time that went into this contribution, and we are also excited to get it merged! Quick update that we're testing it internally and will get back to you with feedback soon.

index.d.ts
```diff
@@ -192,8 +192,12 @@ export interface ITransformEntriesConfig {
...
   */
   transformEntryForLocale: (fromFields: ContentFields, currentLocale: string) => any,
-  /** (optional) – If true, the transformed entries will be published. If false, both will remain in draft state (default true) */
-  shouldPublish?: boolean
+  /** (optional) – If true, the transformed entries will be published. If false, they will remain in draft state. When the value is set to "preserve" items will be published only if the original entry was published as well (default true) */
+  shouldPublish?: boolean|"preserve",
+  /** (optional) – If true, references to the old entry are replaced with references to the new entry (default true) */
+  updateReferences?: boolean,
+  /** (optional) – If true, the original entry is removed after the new entry was created (default true) */
+  removeOldEntries?: boolean
```
TimBeyer: This is not relevant for `transformEntries` so maybe we should have a separate interface for the `transformEntriesToType` config. Same for `updateReferences`.

src/lib/action/entry-transform-to-type.ts
```diff
@@ -0,0 +1,123 @@
...
+            targetEntry.setField(fieldId, {})
+          }
+
+          for (const [locale, localizedValue] of _.entries(localizedField)) {
+            targetEntry.setFieldForLocale(fieldId, locale, localizedValue)
+          }
+
+        }
+        await api.saveEntry(targetEntry.id)
+        if (this.shouldPublish === true || (this.shouldPublish === 'preserve' && entry.isPublished) ) {
```
TimBeyer: Could you also port this behavior to the standard entry transform action?
Otherwise I fear the APIs diverge too much, plus it's a very useful feature to have.

src/lib/offline-api/index.ts
```diff
@@ -341,6 +363,46 @@ class OfflineAPI {
...
+    const hasEntry = this.entries.some((entry) => entry.id === id)
+
+    if (!hasEntry) {
+      throw new Error(`Cannot unpublish Entry ${id} because it does not exist`)
+    }
+
+    const entry = this.entries.find((entry) => entry.id === id)
+
+    // Store clone as a request
+    this.currentRequestsRecorded.push(unpublishEntryRequest(entry.clone()))
```
TimBeyer: In this case we can skip the clone since both requests do not require the entry payload, and `id` and `version` cannot be mutated once the request objects have been created

src/lib/entities/entry.ts
```diff
@@ -1,17 +1,20 @@
...
   private _version: number
   private _fields: object
+  private _isPublished: boolean
 
   constructor (entry: APIEntry) {
     this._id = entry.sys.id
     this._fields = entry.fields
     this._version = entry.sys.version
     this._contentTypeId = entry.sys.contentType.sys.id
+    this._isPublished = !isNullOrUndefined(entry.sys.publishedVersion)
```
TimBeyer: When an entry is cloned, we call the constructor with `this.toApiEntry`, but that does not include the `publishedVersion`. I think that may cause issues, so probably we should also include the `publishedVersion` in the clone.

We should keep in mind to remove `publishedVersion` then after an `unpublish` though, so that the next clone does not restore the published state.

src/lib/offline-api/index.ts
```diff
@@ -341,6 +363,46 @@ class OfflineAPI {
...
+
+    // Mutate version bump
+    entry.version = entry.version + 1
+    entry.isPublished = false
```
TimBeyer: I think in addition to that we need to set `entry.sys.publishedVersion` to `undefined`.

src/lib/offline-api/index.ts
```diff
@@ -341,6 +363,46 @@ class OfflineAPI {
...
+
+    if (!hasEntry) {
+      throw new Error(`Cannot delete Entry ${id} because it does not exist`)
+    }    // Store clone as a request
+    const entry = this.entries.find((entry) => entry.id === id)
+
+    const index = this.entries.indexOf(entry)
+    this.entries.splice(index, 1)
+
+    this.currentRequestsRecorded.push(deleteEntryRequest(entry.clone()))
```
TimBeyer: Same thing about the `clone` above.

src/lib/offline-api/index.ts
```diff
@@ -351,6 +413,33 @@ class OfflineAPI {
     return entries
   }
 
+  async getLinks (childId: string, locales: string[]): Promise<Link[]> {
+
+    const links: Link[] = []
+
+    this.entries.forEach((entry) => {
```
TimBeyer: Code style wise, I think a `for ... of` loop would be preferable here.

src/lib/offline-api/index.ts
```diff
@@ -351,6 +413,33 @@ class OfflineAPI {
     return entries
   }
 
+  async getLinks (childId: string, locales: string[]): Promise<Link[]> {
```
TimBeyer: This is a super expensive operation.
Maybe we should index all links once when we initialize all the entries, then again it would make it more difficult to deal with updates to entries during the migration, which would also have to be added to the index.
I feel like for now this may be good enough, but I can already see a few people with larger spaces having performance problems due to repeatedly iterating every single entry in their entire space to look for links.

aKzenT:
Thanks for the review. We will take a look soon and update the PR where necessary.

src/lib/entities/entry.ts
```diff
@@ -40,6 +43,18 @@ class Entry {
     this._fields[id] = field
   }
 
+  replaceArrayLinkForLocale (id: string, locale: string, index: number, linkId: string) {
+    const link = { sys: { id: linkId, type: 'Link', linkType: 'Entry'}}
+    const field = this._fields[id] || {}
+    const fieldArray = field[locale] as Array<any> || new Array<any>()
+
+    if (fieldArray.length < index + 1) {
```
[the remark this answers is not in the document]
aKzenT: I don't even remember, why we do not simply assign the index. Maybe it was a case of simple thinking to complex. Do you have any concern in assigning it directly?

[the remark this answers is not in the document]
TimBeyer: I'm thinking maybe it could happen that this method isn't called for every index so you'd end up with holes in the array that then mess with things later on?

If that's not the case then I don't see a problem with just assigning it.

src/lib/action/entry-transform-to-type.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  async applyTo (api: OfflineAPI) {
+    const entries: Entry[] = await api.getEntriesForContentType(this.sourceContentTypeId)
+    const locales: string[] = await api.getLocalesForSpace()
+
+    for (const entry of entries) {
+      const inputs = this.fromFields ? _.pick(entry.fields, this.fromFields) : entry.fields
+      const newEntryId = await this.identityKey(inputs)
+      const hasEntry = await api.hasEntry(newEntryId)
```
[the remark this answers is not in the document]
aKzenT: This is a "feature" we borrowed from the regular entry-derive. It gives you the possibility to replace several entries to a single one or to be able to resume a migration if it has failed previously.

But I agree that this might also lead to unexpected behavior, so we can remove it and reintroduce again if necessary as a separate option.

index.d.ts
aKzenT: Makes sense

src/lib/entities/entry.ts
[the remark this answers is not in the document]
aKzenT: But even in this case, I don't see that we would create holes. The splice replaces the item at the given index and we only call the method for indices that already exist (hence replace). @GordonApplepie Any reason you choose splice in this case instead of array assignment?

src/lib/action/entry-transform-to-type.ts
aKzenT: Done

src/lib/offline-api/index.ts
aKzenT: You are correct. However it seems that OfflineApi always clones the entry before creating the requests, so the behavior is consistent with the rest of the code. If we change this, we should change it for all of the requests. What do you think?

TimBeyer: You're correct. Maybe let's just keep it consistent and always clone like you already did.
The costs are most likely negligible and if anything ever changes we're always on the safe side.

src/lib/offline-api/index.ts
aKzenT: See above. Is there any instance where the clone() is really necessary to create the request?

src/lib/offline-api/index.ts
aKzenT: Done. I left the innermost forEach as is since iteration over entries() to get both the key and value is slower that using forEach and there are potentially a lot of loop iterations, so it might be noticeable.

src/lib/offline-api/index.ts
aKzenT: Fully agree. We talked about this but weren't sure what the best method would be to do this. Especially considering that we do not want to penalize users that do not use this type of transform and that we have to make sure to keep this up to date like you said.

Also we thought that the PR is big enough as it is, so it might be better to have this as a separate improvement.

That being said we actually tested the migration on a big space with several thousand items and did not have a problem performance wise so far. Most of the time spent was actually at the beginning since we need to fetch all entries in the space to make this kind of calculation in the first place, which can take some time on a slow network.

aKzenT:
@TimBeyer can you take a look at the changes? We are planning to do several more migrations in the beginning of January, which will further test this new migration type.

Khaledgarbaya:
Hey @aKzenT,

This is ready to merge. can you please run `npm run lint` and fix the linting issues and push again.

After that, I will merge this PR.

GordonApplepie:
happy new year all together. @Khaledgarbaya I fixed the linter errors as requested. I am a bit unsure of the warning "no-unused-variable is deprecated" since this is a linter settings issue. I would ignore this for now?

TimBeyer:
@GordonApplepie you can ignore it for now.  
I think this is now a Typescript setting and so the linter option has been deprecated.

Khaledgarbaya:
Hey @aKzenT and @GordonApplepie we would like to send you some thank you swags for your contributions.
Can you ping me at this email khaled[at]contentful.com with your addresses so we can ship them to you?

GordonApplepie:
Hi @TimBeyer, I will look after that now. Sadly I found a problem due to recent updates in the master branch regarding the "fetcher.ts". There is a new filter implementation that actually prevents to load the complete list of entries which we need to update the references.

I will describe it again in the pull request. Editing the readme file now.

jarkkosyrjala:
I'm getting "Validation failed" error when trying to use `shouldPublish: 'preserve'
`"string" is not a valid type for the content transformation property "shouldPublish". Expected "boolean".`

Full example of the migration file:
```
  const fields = ['field1','field2','field3']
  const contentType = 'exampleContentType'
  const sourceLocale = 'en'
  const targetLocale = 'en-US'
  migration.transformEntries({
    contentType: contentType,
    from: fields,
    to: fields,
    shouldPublish: 'preserve',
    transformEntryForLocale: function (fromFields, currentLocale) {
      if (currentLocale === targetLocale) {
        return fields.reduce(function(result, fieldName) {
          if (fromFields[fieldName] && fromFields[fieldName][sourceLocale]) {
            result[fieldName] = fromFields[fieldName][sourceLocale]
          }
          return result;
        }, {});
      } else {
        // Ignore other locales
        return;
      }
    }
  });
};
```

Should I create a separate Issue about this?

GordonApplepie:
@jarkkosyrjala I think you are mixing two different functions here "transformEntries" is not "transformEntriesToType". "preserve" is only available in the last one.

TimBeyer:
@GordonApplepie if I saw it correctly, then `'preserve'` was also implemented for the normal `transformEntries` in this PR. Probably it just wasn't added to the validations.

GordonApplepie:
#163

TimBeyer:
@GordonApplepie I made a PR that fixes the issue in https://github.com/contentful/contentful-migration/pull/164

jarkkosyrjala:
> @GordonApplepie if I saw it correctly, then `'preserve'` was also implemented for the normal `transformEntries` in this PR. Probably it just wasn't added to the validations.

Made a PR that fixes this: https://github.com/contentful/contentful-migration/pull/165

PR #151, opened 2018-11-27
fix(validations): do not require min and max size validation

phoebeschmidt:
Relax migration tool validations for min/max size field validations. Now they are not required, but still nullable.


test/end-to-end/validations.spec.js
```diff
@@ -13,7 +13,7 @@ const ENVIRONMENT_ID = uuid.v4();
 
 const SOURCE_TEST_SPACE = process.env.CONTENTFUL_INTEGRATION_SOURCE_SPACE;
 
-describe('apply validations migration examples', function () {
+describe.only('apply validations migration examples', function () {
```
Khaledgarbaya:
```suggestion
describe('apply validations migration examples', function () {
```

PR #152, opened 2018-11-27
fix(validation): add validation schema for rich text nodes

phoebeschmidt:


src/lib/offline-api/validator/schema/field-validations-schema.ts
```diff
@@ -31,6 +31,32 @@ const assetImageDimensions = validation('assetImageDimensions', Joi.object({
 
 const assetFileSize = validation('assetFileSize', range('number'))
 
+const nodes = validation('nodes', Joi.object({
+  'embedded-entry-block': Joi.array(),
```
phoebeschmidt: here decided to be very general, although AFAIK these are the possible validations here:
(object with `size` and `message`, object with `linkContentType` and `message`
```
{"nodes": {
	"embedded-entry-inline": [
              {
                "size": {
                  "min": 10,
                  "max": 20
                },
                "message": "this is a custom error for number of embedded inline entries"
              },
              {
                "linkContentType": [
                  "parent"
                ],
                "message": "we only accept parent"
              }
    ]
}
```

do you think I should go one level deeper with possible validations here?

Issue #122, opened 2018-07-19
Migration unsuccessful with transformEntries because "Cannot edit archived" and "Cannot publish archived"


nicco91:
I'm trying to do a migration which create a fields and then transform entries creating values for the new field created before.

## Expected Behavior

Run the migration, create the field, fill it with the values I create.

## Actual Behavior

Field is created correctly, values are generated correctly, but the migration fails because it try to publish an entry which is in status `archived`.

```
$ node migrate.js --portal tcom --env test --migration migrations/3.13.0.js
The following migration has been planned

Update Content Type pack

  Create field servicesDescription
    - name: "Services Description"
    - type: "Text"

Publish Content Type pack
Transform entries for pack
  - from: services
  - to: servicesDescription
? Do you want to apply the migration Yes
 √ Update Content Type pack
 > Transform entries for pack
   × Making requests (48/48)
     → Batch failed
🚨  Migration unsuccessful
Batch failed

Error: {"status":"Bad Request","message":"Cannot edit archived","details":{},"url":"https://api.contentful.com:443/spaces/fzchbmaavok4\\environments\\develop\\entries\\2iqxnK1F1aakq24U4C8CSq"}


Error: {"status":"Bad Request","message":"Cannot publish archived","details":{},"url":"https://api.contentful.com:443/spaces/fzchbmaavok4\\environments\\develop\\entries\\2iqxnK1F1aakq24U4C8CSq\\published"}
```

I have no control on the entries status with the `transformEntries` migration function. I just provide the `transformEntryForLocale` function.
So, how can I run this migration skipping the archived entries?
It actually works for the entries that are not `archived` but the error message is shown and it says "migration unsuccessful", even if it is just partially unsuccessful.

## Possible Solution

Parameters on the `transformEntries` function for choose entries status or just skip the archived entries and return a success message. 

## Steps to Reproduce

Run any migration with `transformEntries` function where there are some `archived` entries.

## Environment

-   **Node Version**: 8.11.3
-   **Package Manager Version**: 5.6.0
-   **Operating System**: Windows 10 64-bit
-   **Package Version**: 0.10.0

stoltzrobin:
@Khaledgarbaya have there been any progress on this?

Khaledgarbaya:
This should be fixed by https://github.com/contentful/contentful-migration/pull/143

TimBeyer:
I think this will need a separate fix from #143 
[...]

Issue #108, opened 2018-06-18
Cannot edit archived


jarkkosyrjala:
## Expected Behavior

Migration should edit or skip archived entries

## Actual Behavior

Migration fails with "status":"Bad Request" and "message":"Cannot edit archived"

## Possible Solution

Just skip archived entries

## Context

Doing a transformEntryForLocale (copying content from one locale to another)

jarkkosyrjala:
This appears to be the same bug as https://github.com/contentful/contentful-migration/issues/122

PR #161, opened 2019-01-03
fix: do not fetch archived entries

TimBeyer:
Since we cannot apply any transformations on archived entries, and since attempting to update archived entries results in an error, we skip fetching those from the API.

They will however still be taken into account when trying to delete a content type, so that we do not run into problems with validations there.

Fixes https://github.com/contentful/contentful-migration/issues/122
Fixes https://github.com/contentful/contentful-migration/issues/108


test/unit/lib/fetcher.spec.ts
```diff
@@ -118,6 +118,7 @@ describe('Fetcher', function () {
       })
     ]
     const fetcher = new Fetcher(request)
+    console.log('Fetchering')
```
phoebeschmidt: remove console log pls :)

PR #163, opened 2019-01-04
TransformEntriesToType: readme extended, fixed small issue

GordonApplepie:
@TimBeyer as requested I extended the `README.md`. You will find the problem with the `fetcher.ts` also in this PR.

The problem is that we need the entries of all content types since we are tryíng to update the parental entries and replace the old linked entries to the newly created ones.


TimBeyer:
Could you separate the PRs so that we have one clean PR branched from master that does the readme change, and then we can fix the ID filter separately.

GordonApplepie:
@TimBeyer yeah. true that was not so clever actually. I also saw that it is fixed already 😃 will close down this PR and create a new one for the readme file.

PR #165, opened 2019-01-04
fix ContentTransformIntentValidator

jarkkosyrjala:
## Summary

shouldPublish should accept 'preserve' in addition to boolean values

## Description

Add 'preserve' as a valid option to validations for shouldPublish field

## Motivation and Context

https://github.com/contentful/contentful-migration/pull/143#issuecomment-451451263

### DIFF
--- a/test/unit/lib/intent-validator/content-transform.spec.ts
+++ b/test/unit/lib/intent-validator/content-transform.spec.ts
@@ -157,7 +157,7 @@ describe('Content transformation', function () {
                 type: 'contentType/transformEntries'
               }
             },
-            message: '"undefined" is not a valid type for the content transformation property "shouldPublish". Expected "alternatives".',
+            message: '"yes please" is not a valid value for the content transformation property "shouldPublish". Expected type boolean or string value "preserve".',
             type: 'InvalidType'
           }
         ])
--- a/test/unit/lib/intent-validator/content-transform.spec.ts
+++ b/test/unit/lib/intent-validator/content-transform.spec.ts
@@ -157,7 +157,7 @@ describe('Content transformation', function () {
                 type: 'contentType/transformEntries'
               }
             },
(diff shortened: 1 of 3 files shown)


TimBeyer:
I guess we also need to add this for `EntryTransformToTypeIntentValidator` then?

TimBeyer:
And actually `EntryTransformToTypeValidator` needs to check for `isEntryTransformToType` instead of `isEntryDerive`. 

Must have missed this during the review.

jarkkosyrjala:
@TimBeyer Also, this makes the content-transform.spec.ts test to fail, because the error message from the validation is now different at https://github.com/contentful/contentful-migration/blob/0203edf5267b07450ae0cb73e37660d5c456185f/test/unit/lib/intent-validator/content-transform.spec.ts#L160

test/unit/lib/intent-validator/content-transform.spec.ts
```diff
@@ -157,7 +157,7 @@ describe('Content transformation', function () {
                 type: 'contentType/transformEntries'
               }
             },
-            message: '"string" is not a valid type for the content transformation property "shouldPublish". Expected "boolean".',
+            message: '"undefined" is not a valid type for the content transformation property "shouldPublish". Expected "alternatives".',
```
TimBeyer: This new error message isn't really useful anymore for people.  
Maybe we could figure out in https://github.com/contentful/contentful-migration/blob/0203edf5267b07450ae0cb73e37660d5c456185f/src/lib/intent-validator/schema-validator.ts#L78 whether we are dealing with `Joi.alternatives` and figure out the different types that are allowed.  
Either that, or somehow change the message for the `alternatives` validation so that it doesn't say `Expected "alternatives"`.

jarkkosyrjala: @TimBeyer error message fixed with tests

TimBeyer:
👏

PR #167, opened 2019-01-05
Make typings compatible with strict typescript

geigerzaehler:
Before, TypeScript projects using the `strict` compiler settings were unable to compile when `contentful-migration` was imported. The following error was displayed

    error TS7010: 'toTheTop', which lacks return-type annotation, implicitly has an 'any' return type.


geigerzaehler:
Done

PR #170, opened 2019-03-04
add `unpublish` to `shouldPublish` option of `transformEntry` function

cachrisman:
## Summary

This PR adds the option to unpublish entries that have been processed by the `transformEntry` function

## Description

In addition to `true` and `false` as options for the `shouldPublish` configuration option, this adds a `unpublish` option that makes an api call to unpublish the entry being transformed.

## Motivation and Context

This is for the situation where the validations of one or more content types are changed such that existing entries would now fail validations and not be publishable without updates. If you are releasing a new version of your frontend and know that only entries which pass the updated content model validation changes will work, this is a way to show your content creators and editors which entries need to be updated to work with the updated frontend. 

I did this because of a question in the Contentful Community slack: https://contentful-community.slack.com/archives/CDY4JBMTJ/p1551255690020000

## Todos

- This could potentially be spun out from the `shouldPublish` option into a `validateEntry` function that checks entries to see if they would pass validation and unpublishes only the ones that fail the revalidation.
- Might also want to add this similar functionality to the `deriveLinkedEntries` and `transformEntriesToType` functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ The transform function is expected to return an object with the desired target f
     - `fields` is an object containing each of the `from` fields. Each field will contain their current localized values (i.e. `from == {myField: {'en-US': 'my field value'}}`)
     - `locale` one of the locales in the space being transformed
   The return value must be an object with the same keys as specified in `to`. Their values will be written to the respective entry fields for the current locale (i.e. `{nameField: 'myNewValue'}`). If it returns `undefined`, this the values for this locale on the entry will be left untouched.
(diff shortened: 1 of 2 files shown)


Gi11i4m:
I'd definitely focus on the _validateEntry_ function. Not sure if this wouldn't collide with the _shouldPublish_ option. It would be very convenient to be able to automatically unpublish entries that are no longer valid after transformation. 
Otherwise an application assuming that the content from contentful adheres to the content-model could fail after a migration, since invalid published entries are possible this way.

Issue #173, opened 2019-03-19
running migration resets custom sidebar 


formatlos:
I'm using a custom sidebar for a content-type, but after running any migration on that content-type (e.g. adding a field) the sidebar gets reset to the default one. 

This is quite odd, as my migrations run automated and because of this issue theres an additional manual step to re-configure the sidebar. 

Is there any way to avoid this?

suevalov:
@formatlos 
It's a known issue and we're planning to fix the issue this week. Sorry for inconvenience and thanks for bringing this up!

PR #175, opened 2019-03-26
fix: editorInterface methods do not reset custom sidebar anymore

suevalov:
## Summary
 
Fixes https://github.com/contentful/contentful-migration/issues/173

## Description

`editorInterface` payload can have custom sidebar configuration saved. Currently all related methods just override `sidebar` configuration. This PR fixes the behavior, so if `sidebar` configuration isn't overwritten if present in the initial payload.


Khaledgarbaya:
@suevalov can you add tests to the editor interfaces intent at least to make sure that the tool gives an accurate summary about what will happen

[here](https://github.com/contentful/contentful-migration/blob/master/test/unit/lib/intent/editorinterface-update.spec.ts) for example

suevalov:
@Khaledgarbaya I think it's quite accurate at the moment. In this PR we're addressing the problem that the tool does what is not intended to happen (rewrites `sidebar` config).

There will be an another PR which adds possibility to `copy` and `create` sidebar configuration.

Issue #179, opened 2019-04-24
Using transformEntriesToType after changeFieldId causes reference error


yanneves:
## Expected Behavior

Using `.changeFieldId()` in the same migration as `.transformEntriesToType()` should work as defined.

## Actual Behavior

When `.changeFieldId()` is used before `.transformEntriesToType` in a migration, a reference error is thrown by the `getLinks()` method in `lib/offline-api/index.js` attributed to this line:

```typescript
const field = entry.fields[key][locale] // entry.fields[key] is undefined
```

## Possible Solution

Only use one of the aforementioned methods in a migration script at a time

## Context

Running a custom migration to merge two content types

## Environment

-   **Node Version**: v11.14.0
-   **Package Manager Version**: 6.9.0
-   **Operating System**: Linux x64
-   **Package Version**: contentful-migration@0.17.2

yanneves:
Fixed in #180 and released with version 0.17.4 🎉 

Thanks @Khaledgarbaya and @phoebeschmidt!

PR #180, opened 2019-04-24
fix: Safer reads on entry fields properties inside getLinks() method for Offline API

yanneves:
Follows #179

### DIFF
--- a/src/lib/offline-api/index.ts
+++ b/src/lib/offline-api/index.ts
@@ -1,4 +1,4 @@
-import { omit, compact } from 'lodash'
+import { omit, compact, get } from 'lodash'
 import FieldDeletionValidator from './validator/field-deletion'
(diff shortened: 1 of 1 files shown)


yanneves:
I notice forked branches in PRs always fail, because they miss settings defined in the repository (since the fork is based in a new repository):

```bash
Setting environment variables from repository settings
$ export CODECOV_TOKEN=[secure]
$ export GH_TOKEN=[secure]
$ export CONTENTFUL_INTEGRATION_MANAGEMENT_TOKEN=[secure]
$ export CONTENTFUL_INTEGRATION_SOURCE_SPACE=[secure]
$ export NPM_TOKEN=[secure]
```

Would it make sense to change these environment variables to be defined as project settings in Travis CI,  rather than repository settings?

Khaledgarbaya:
Thanks @yanneves

PR #181, opened 2019-04-30
editorInterface methods do not reset custom editor extension

suevalov:
## Summary

Fix for a new custom editor extensions feature.

## Description

Change in editorInterface implementation so migrations don't override `editor` property if it was set already.

### DIFF
--- a/src/lib/entities/content-type.ts
+++ b/src/lib/entities/content-type.ts
@@ -152,9 +152,9 @@ class EditorInterfaces {
     })
 
     const result: {
-      controls: any,
-      sidebar?: any,
-      editor?: any
+      controls: APIEditorInterfaceControl[],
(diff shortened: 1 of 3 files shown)


src/lib/entities/content-type.ts
```diff
@@ -140,23 +142,31 @@ class EditorInterfaces {
...
         fieldId: c.fieldId,
         widgetId: c.widgetId,
         settings: c.settings
       })
     })
+
+    const result: {
+      controls: any,
+      sidebar?: any,
+      editor?: any
```
Khaledgarbaya: it would be awesome if we could be more specific here instead of `any` I think we already have `APIEditorInterface...` types that we can probably use, I might be wrong though

PR #182, opened 2019-04-30
chore: "Usage as a library" example error logging

fsarachu:
## Summary

Fix "Usage as a library" example error logging

## Description

Now pass the error object to `console.error`

## Motivation and Context

That example was wrong, so I'm fixing it.


Khaledgarbaya:
Thanks @fsarachu

PR #192, opened 2019-06-06
feat(editorInterface): Add support for sidebar and entry editor extensions

lehnerm:
## Summary

Adds support for sidebar and entry editor configuration in a ContentType's EditorInterface.  

Sidebar methods added:
- `addSidebarWidget`
- `removeSidebarWidget`
- `addSidebarWidget`

Control methods changed:
- `updateFieldControl` was `updateEditorInterface (deprecated)`
   - added support for widgetNamespace and extension settings
- `resetFieldControl` was `resetEditorInterface (deprecated)`
- `copyFieldControl` was `copyEditorInterface (deprecated)`

Entry Editor methods added:
- `setEntryEditor`
- `resetEntryEditorToDefault()`


index.d.ts
```diff
@@ -140,29 +148,63 @@ export interface ContentType {
...
-   * Changes the editor interface of given field's ID.
+   * Changes the control of given field's ID.
    *
    * @param fieldId The ID of the field.
    * @param widgetId The new widget ID for the field.
    * @param settings Widget settings
+   * @param widgetNamespace The namespace of the widget. Use 'builtin' for standard widgets or 'extension' for UI extensions.
    */
-  changeEditorInterface (fieldId: string, widgetId: string, settings?: IEditorInterfaceOptions): void
+  changeFieldControl (fieldId: string, widgetId: string, settings?: IEditorInterfaceOptions, widgetNamespace?: 'builtin' | 'extension'): void
```
Khaledgarbaya: I think this is a big breaking change and we should keep the old methods as an alias to the new ones and we log a deprecation message to give the users some time to change their code and in another release we remove them entirely. WDYT?

lehnerm: Yes, the actual implementation still keeps them around; how would you deprecate them?

Khaledgarbaya: we can add a console.log('(deprecated) changeEditorInterface is deprecated and will be removed in a later version please use changeControl instead').
and since we are using typescript we can use something similar to [this](https://gist.github.com/michelsalib/cfff8114ce081930a453) `@deprecated('message')` tag but I am not sure if it's stable enough

lehnerm: Opted to keep it simple and add a warning as well as @deprecated in the jsdoc of the index.d.ts file

README.md
```diff
@@ -75,9 +75,13 @@
       - [`editField(id[, opts])` : [Field](#field)](#editfieldid-opts--fieldfield)
       - [`deleteField(id)` : void](#deletefieldid--void)
       - [`changeFieldId (currentId, newId)` : void](#changefieldid-currentid-newid--void)
-      - [`changeEditorInterface (fieldId, widgetId[, settings])` : void](#changeeditorinterface-fieldid-widgetid-settings--void)
-      - [`resetEditorInterface (fieldId)` : void](#reseteditorinterface-fieldid--void)
-      - [`copyEditorInterface (sourceFieldId, destinationFieldId)` : void](#copyeditorinterface-sourcefieldid-destinationfieldid--void)
+      - [`changeFieldControl (fieldId, widgetId[, settings, widgetNamespace])` : void](#changefieldcontrol-fieldid-widgetid-settings-widgetnamespace--void)
```
jelz: Can the signature be `changeFieldControl(fieldId, widgetNamespace, widgetId[, settings])`?

It's more logical (widget lives in a namespace, not the other way around) and also makes it mandatory which is not enforced by the API (it would be breaking change) but in the tooling should be enforced.

README.md
```diff
@@ -75,9 +75,13 @@
...
+      - [`resetFieldControl (fieldId)` : void](#resetfieldcontrol-fieldid--void)
+      - [`copyFieldControl (sourceFieldId, destinationFieldId)` : void](#copyfieldcontrol-sourcefieldid-destinationfieldid--void)
+      - [`addSidebarWidget (widgetId[, widgetNamespace, insertBeforeWidgetId, settings])` : void](#addsidebarwidget-widgetid-widgetnamespace-insertbeforewidgetid-settings--void)
```
jelz: The same: `widgetNamespace` in front. Do we put the widget at the end if `insertBeforeWidgetId` is undefined?

README.md
```diff
@@ -75,9 +75,13 @@
...
+      - [`updateSidebarWidget (widgetId, settings)` : void](#updatesidebarwidget-widgetid-settings--void)
```
jelz: Right now there are no settings for built-in widgets but sooner or later they will be added. Then we need both ID and namespace to address a widget to be changed.

README.md
```diff
@@ -75,9 +75,13 @@
...
+      - [`removeSidebarWidget (widgetId)` : void](#removesidebarwidget-widgetid--void)
```
jelz: For this one we need the namespace for sure. There may be a widget with the same ID in multiple namespaces.

README.md
```diff
@@ -517,15 +521,15 @@ module.exports = function (migration) {
...
 
-Changes the editor interface of given field's ID.
+Changes control interface of given field's ID.
 
 **`fieldId : string`** – The ID of the field.
 
 **`widgetId : string`** – The new widget ID for the field. See the [editor interface documentation](https://www.contentful.com/developers/docs/concepts/editor-interfaces/) for a list of available widgets.
 
-**`settings : Object`** – Widget settings, with the following options:
+**`settings : Object`** – Widget settings and extension instance parameters. For builtin widgets, the the following options are available:
```
jelz: I'd define that for instance parameters `settings` are a free-form object with values being numbers, booleans or strings.

README.md
lehnerm: Well the argument order is based on the old methods to make the sig compatible

README.md
lehnerm: as outlined in the docs, yes

README.md
lehnerm: because of ID collisions, or what is the concern?

README.md
lehnerm: According to alex they could be objects as well, therefore they are defined as any. is this incorrect?

README.md
jelz: Yeah but we introduce a new method so it's a chance for us to make things better?

README.md
lehnerm: Extension ID's are some form of UUIDs or? And builtin stuff is a descriptive string. How are we going to make them collide?

README.md
lehnerm: changed it

README.md
jelz: You can PUT to `/extensions/:id` to set your own ID if the entity doesn't exist yet.

It's actually the default behavior of the Contentful CLI (uses `id` property from `extension.json`)

lehnerm: I'll change it

README.md
suevalov: let's double check API schema. I might be wrong

README.md
```diff
@@ -535,14 +543,54 @@ Changes the editor interface of given field's ID.
...
 
 **`fieldId : string`** – The ID of the field.
 
-#### `copyEditorInterface (sourceFieldId, destinationFieldId)` : void
+#### `copyFieldControl (sourceFieldId, destinationFieldId)` : void
+
+**`sourceFieldId : string`** – The ID of the field to copy the control setting from.
+**`destinationFieldId : string`** – The ID of the field to apply the copied control setting to.
+
+#### `addSidebarWidget (widgetId, widgetNamespace[, insertBeforeWidgetId, settings])` : void
```
suevalov: I'd swap `insertBeforeWidgetId` and `settings`. It would seem more aligned with other methods and won't require to set `insertBeforeWidgetId` to null if you want to skip it.

README.md
```diff
@@ -535,14 +543,54 @@ Changes the editor interface of given field's ID.
...
 
 **`fieldId : string`** – The ID of the field.
 
-#### `copyEditorInterface (sourceFieldId, destinationFieldId)` : void
+#### `copyFieldControl (sourceFieldId, destinationFieldId)` : void
+
+**`sourceFieldId : string`** – The ID of the field to copy the control setting from.
+**`destinationFieldId : string`** – The ID of the field to apply the copied control setting to.
+
+#### `addSidebarWidget (widgetNamespace, widgetId[, insertBeforeWidgetId, settings])` : void
```
suevalov: I see `disabled` property at actual implementation, but it's not listed here.

README.md
lehnerm: https://github.com/contentful/content_api/blob/3e4bf6b4a7f648287fa8da3236b55c447190efb9/lib/handler/constraints/management/schemas/shared.js#L35

This states number, boolean or string

README.md
```diff
@@ -535,14 +545,66 @@ Changes the editor interface of given field's ID.
...
+- `sidebar-builtin` (Standard widget, default)
+- `extension` (Custom UI extension)
+
+**`widgetId : string`** – The ID of the builtin or extension widget to remove.
+
+#### `resetSidebarToDefault ()` : void
+
+Resets the sidebar of the content type to default.
+
+#### `configureEntryEditor (widgetNamespace, widgetId[, settings])** : void
```
jelz: Backtick missing

PR #194, opened 2019-06-15
fix: TypeScript declaration type checks in strict mode

geigerzaehler:
We update `index.d.ts` so that it type checks when `--strict` is passed to the TypeScript compiler.

We also add a test that ensures that this does not break again in the future.


Khaledgarbaya:
Thanks @geigerzaehler

Issue #198, opened 2019-07-01
changeFieldControl does not seem to work correctly with custom extensions


ThomasJacobsAppno:
Hi,

I tried using changeFieldControl within a migration script to change the control to a custom extension (and apply settings). In the below example I tried using the "Bynder Assets" plugin from the marketplace, but I have also attempted this with my own custom plugin.

`element.changeFieldControl('image', 'bynder-assets', 'extension', { bynderURL: 'URL_VALUE'});`

When I check the content element afterwards, the control isn't set to any widget (not even the default one). The field was created like this:

`element.createField('image').name('Image').type('Object').localized(true).required(false).disabled(false).omitted(false)`

The strange part is, that the "bynderURL" property was set correctly on the widget, so this seems to be an issue with the migration itself.

Thanks in advance.

Khaledgarbaya:
Hi @ThomasJacobsAppno,
could that be related to the [issue](https://github.com/contentful/contentful-migration/pull/199) we had in the documentation?

ThomasJacobsAppno:
@Khaledgarbaya, I noticed that. It is possible. Although the method parameters are now the same as in the documentation (and my example above).

I just checked this further and I may have found the issue in https://github.com/contentful/contentful-migration/blob/master/src/lib/migration-steps/index.ts at line 148. The parameters are still in a different order there.

I am not sure which order should be correct, but they should definitely be the same (I would think).

lehnerm:
@ThomasJacobsAppno you are correct, the argument order between docs and actual implementation is inconsistent. Your usage, assuming `bynder-assets` is the extension id, is correct for the current version though.

Is `bynder-assets` the actual id of the bynder extension installed in your space?  Please note that the marketplace installer does not give the extension a fixed named id, rather it will create a random one on installation. Can you please double check the id for your bynder installation?

Additionally, I just ran the following example after installing `bynder-assets` from the marketplace:
```
module.exports = function (migration) {
  const bynderExtensionId = '1X1j*********************'
  const contentType = migration.createContentType('bynder_demo', {
    name: 'Bynder Demo'
  });

  contentType.createField('asset', {
    name: 'Asset',
    type: 'Object',
    required: true
  });

  contentType.changeFieldControl('asset', bynderExtensionId, 'extension', {
    bynderURL: 'https://www.example.com'
  });
};
```
Content type and field are created as well as the correct plugin assigned to it.

ThomasJacobsAppno:
[...]

PR #203, opened 2019-07-10
Consistent signature of changeFieldControl() in README, index.d.ts and index.ts

lehnerm:
## Summary

Fixes inconsistency in changeFieldControl signature raised in issue #198

Same as for the sidebar methods, widgetNamespace should always be mandatory and always precede widgetId. This is due to the fact that widgetId is only unique in said namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@
       - [`editField(id[, opts])` : [Field](#field)](#editfieldid-opts--fieldfield)
       - [`deleteField(id)` : void](#deletefieldid--void)
       - [`changeFieldId (currentId, newId)` : void](#changefieldid-currentid-newid--void)
-      - [`changeFieldControl (fieldId, widgetId, widgetNamespace[, settings])` : void](#changefieldcontrol-fieldid-widgetid-widgetnamespace-settings--void)
+      - [`changeFieldControl (fieldId, widgetNamespace, widgetId[, settings])` : void](#changefieldcontrol-fieldid-widgetnamespace-widgetid-settings--void)
       - [`resetFieldControl (fieldId)` : void](#resetfieldcontrol-fieldid--void)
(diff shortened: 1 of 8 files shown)


lehnerm:
> Thanks for this, @lehnerm.
> 
> A couple of questions I have:
> Any reason we're switching the positions of `widgetId` and `widgetNamespace` vs leaving it the way it is? We'll just need to update the documentation/readme if we left it this way it is. I could be missing.
> 
> Would this be a breaking change given this method is already being used with `widgetId` coming before `widgetNamespace`?

This was missed in the initial implementation, and one of the documentation files was later altered to reflect the wrong underlying signature. All other methods using `widgetNamespace` and `widgetId` have the same argument order.

What do you think? API consistency over backwards compatibility?

makinwab:
> 
> This was missed in the initial implementation, and one of the documentation files was later altered to reflect the wrong underlying signature. All other methods using `widgetNamespace` and `widgetId` have the same argument order.
> 
> What do you think? API consistency over backwards compatibility?

I get it now and good question 🤔. Don't know what the right answer is for this case.
I guess if the change introduced to re-order the parameter list in the doc was reverted then it tells me we wanted to go for consistency 🤷🏽‍♀️.

Khaledgarbaya:
> What do you think? API consistency over backward compatibility?

For now, IMO we should go for backward compatibility

ruhkopf:
Hi there, is there an update on this issue? Seems like typescript migrations using changeFieldControl are currently broken. For example the following produces wrong results: `myType.changeFieldControl('myField', 'builtin', 'singleLine', {})` produces 
```
Update field myField
    - widgetId: "builtin"
    - widgetNamespace: "singleLine"
```

For now I'm doing `myType.changeFieldControl('myField', 'singleLine' as any, 'builtin', {})` but obviously this is a confusing workaround...

PR #206, opened 2019-07-18
Provide Management API client instance to migration functions

mc-jordan-andree:
## Description

I'm working on a migration that needs to modify Entry data in Contentful that would leverage the Management API. Since there is a client instance already created in the cli runner, it makes sense to provide the client instance to migration functions so that tasks can be performed.

Also providing the `--apply` cli argument value as well so that different logic can occur when the confirmation dialogue is skipped


mc-jordan-andree:
I need to fix a few things with this since the Typescript transpiler broke. Will re-open a new PR.

PR #213, opened 2019-07-25
Updated typings

zofiag:
Summary
Add missing sourceContentType and targetContentType in interface of transformEntriesToType params + enable `undefined` as a return value in `deriveEntryForLocale`.

Description
contentType: string; was replaced with

sourceContentType: string;
targetContentType: string;

`undefined` was added as accepted return value in `deriveEntryForLocale`.
```
deriveEntryForLocale: (
    inputFields: ContentFields,
    locale: string
  ) => { [field: string]: any } | undefined;
```

Motivation and Context
Make current functionalities consistent with the typings.


index.d.ts
```diff
@@ -212,122 +247,137 @@ export interface ContentType {
...
 
 export interface ITransformEntriesToTypeConfig {
   /** (required) – Content type ID */
-  contentType: string,
+  sourceContentType: string;
+  /** (required) – Content type ID */
+  targetContentType: string;
   /** (required) – Array of the source field IDs */
-  from: string[],
+  from: string[];
```
chrislambe: Since `from` is marked as optional in the [docs](https://github.com/contentful/contentful-migration/blob/master/README.md#transformentriestotypeconfig):

```suggestion
  from?: string[];
```

chrislambe:
Not that I necessarily disagree with your styling changes, but it seems like a pretty sure way to get this PR rejected/ignored. I would love to see this change incorporated since the current interface is wildly incorrect compared to the docs.

zofiag:
sorry for the automated formatting, is now gone. The linked interface `TransformEntryToType` could work if we also enable `string` as a return type from `identityKey`, so
`identityKey: (fromFields: any) => Promise<string> | string`. Then we could import that and reuse in `index.d.ts` for consistency reasons

pappkamerad:
what is the status with that? we would also need correct typings.

PR #218, opened 2019-08-12
fix(typing): 'axios' import in typings

geigerzaehler:
We correctly import `axios` in the type definitions `index.d.ts`.

To ensure that typings are checked on CI we run `npm run test-typescript-declaration`. This requires us to upgrade typescript to the latest version.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-import axios from 'axios'
+import * as axios from 'axios'
 
 export interface Movement {
   toTheTop(): void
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -1,4 +1,4 @@
-import axios from 'axios'
+import * as axios from 'axios'
```
Khaledgarbaya: I might be wrong but this might not necessary since we are using `esModuleInterop`

geigerzaehler: If you revert this change and run `tsc --strict index.d.ts` the compiler will complain about `axios` not being a namespace. This is also why I added the typechecking to CI. `esModuleInterop` only changes how code is imported at runtime.

geigerzaehler:
@Khaledgarbaya Any update on this?

PR #222, opened 2019-09-23
(fix): Resolve error when there are more than one sidebar changes and…

matthew-contentful:
… no content-type changes.

## Summary
A migration with an existing content type and two or more editor entries would fail:

```
module.exports = migration => { 
 const itemEntry = migration.editContentType('item');
 itemEntry.removeSidebarWidget('sidebar-builtin', 'incoming-links-widget'); 
 itemEntry.addSidebarWidget('extension', 'static-ui-extension-incoming-links'); 
};
```
This change detects that groups of editor entry updates to do not need an overall group save operation, avoiding the error condition.


test/integration/migration.spec.js
```diff
@@ -570,4 +571,31 @@ describe('the migration', function () {
       }
     ]);
   }));
+
+  it('adds sidebar widgets to the editor interface of an existing content type', co(function * () {
+    yield migrator(addSidebarWidgetsToExisting);
```
Khaledgarbaya: Any reason we are not using normal `async` `await` here

zcei:
> So now an editor interface change will always also trigger saving the related CT, which may not work in all cases.

What error conditions could happen here apart from network stuff? I wouldn't bother to support stuff like "my content type is invalid but I want to update my editor interface" as both are very tightly coupled from their sheer existence.

> maybe you remember where this came from initially and if our assumptions are correct that it is safe not to always force publish and saving on composed intents?

From what I can recall we argued that a composed intent is a terminating action.
A single intent might be marked to not need a save as it's compose-able, but when you have a set of non-saving intents composed together the whole composition should save after execution.

I do agree that this was focussed on content types initially and the editor interfaces are rather an afterthought, so the naming is slightly off here. But as editor interface changes are nested under the content type, technically it's correct 🤔 

As it does not seem to break any existing tests i wouldn't mind delegating the `shouldSave` and `shouldPublish` to the underlaying intents. In fact, it actually feels cleaner to me rather than hardcoding `true`.

TimBeyer:
My only concern with the delegation would be whether we're missing some sort of terminating saves or publishes now since almost no intent actually defines a `shouldSave` explicitly, but it is the default of the intent base class.
But we have test coverage and I assume those tests do check whether all the operations are leaving things in the correct save or publish state, so probably there's no need to be concerned about this.

PR #229, opened 2019-10-25
fix: shouldPublish preserve no longer publishes "Changed" status entries

### LINKED ISSUE #223: Support `shouldPublish: 'preserve'` option in transformEntries
Transform functions supply the field `shouldPublish` to set whether or not a transformed entry should be published. `transformEntries` supports `boolean` only, while transform entry type supports `bool | 'preserve'`.

**I would like** `transformEntries` to support 'preserve' as well.

**Use case** I am creating a field `A` and want to set its value to the value of field `B`. If I use `shouldPublish: true` my migration will fail for drafts that don't set required fields. If I use `shouldPublish: false` then my migration works, but I have changed every effected entry to `changed` status and would have to publish each one again. 
Thus I would like them to keep their publishing state, if it was drafted stay drafted and vise versa.

loremipson:
## Summary

Should help resolve #223. `shouldPublish` set to `preserve` should not publish entries that are in "Changed" status.

## Motivation and Context

If a Contentful user has made changes to a Published entry that are "Saved" but not "Published", `preserve` currently publishes the entry. This causes issues because the entry might not be ready, or might be missing fields that are marked as required, causing the migration to fail.

## Todos

-   [x] Works with `entry-transform`
-   [x] Works with `entry-transform-to-type`

### DIFF
--- a/src/lib/action/entry-transform-to-type.ts
+++ b/src/lib/action/entry-transform-to-type.ts
@@ -90,8 +90,9 @@ class EntryTransformToTypeAction extends APIAction {
         }
 
       }
+      const currentlyChanged = entry.isChanged
       await api.saveEntry(targetEntry.id)
-      if (this.shouldPublish === true || (this.shouldPublish === 'preserve' && entry.isPublished) ) {
+      if (this.shouldPublish === true || (this.shouldPublish === 'preserve' && (entry.isPublished && !currentlyChanged)) ) {
         await api.publishEntry(targetEntry.id)
       }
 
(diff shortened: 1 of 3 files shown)


loremipson:
Going in to add support for `entry-transform-to-type`. This is a bit different, because the entry is _new_, there's not really a way to get the "Changed" state. And we shouldn't try to "Publish" the current state if there are other changes applied that aren't ready.

I have it set up right now that if `preserve` is set on `transformToType` and one of the entries is "Changed", the new entry is stored in "Draft" state. I believe this works, but I don't know how sufficient it will be down the line, or how much sense it makes in this situation.

PR #231, opened 2019-11-01
fix(stable-results): Order results to protect against changing result…

matthew-contentful:
… set ordering. Increase page size for better performance


src/lib/fetcher.ts
```diff
@@ -151,6 +151,8 @@ export default class Fetcher implements APIFetcher {
 
     while (true) {
       const paramsWithSkip = {
+        limit: '500',
```
TimBeyer: Maybe let's factor this out into a constant so we can adapt this in the future without having to manually change all these tests another time

PR #235, opened 2019-12-04
docs(README): Update README to include "Entry title" phrase

heyitstowler:
## Summary

Updates the README to include the web-app description of a field to the corresponding content model field

## Description

This adds the phrase "Entry title" to the description for the `displayName` option on the Content Model documentation, since that is what a web-app user might know the field as

## Motivation and Context

Just an instance where the wording differs between the web app and the under-the-hood schema. I think adding this kind of clarification can help bridge the knowledge gap for anyone used to using the web app GUI

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ Creates a content type with provided `id` and returns a reference to the newly c
 
 - **`name : string`** – Name of the content type.
 - **`description : string`** – Description of the content type.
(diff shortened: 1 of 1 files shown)


phoebeschmidt:
Thanks @heyitstowler, we're not great about aligning our repos with the web app language, so this is a great improvement!

PR #242, opened 2020-01-02
fix(deletes): stop resending deleted fields

ruderngespra:


test/unit/lib/offline-api/validation/payload-validation-display-field.spec.ts
```diff
@@ -66,7 +66,7 @@ describe('payload validation (display field)', function () {
         [
           {
             type: 'InvalidAction',
-            message: 'Cannot delete field "dessert" on content type "lunch" because it is set as the display field'
```
TimBeyer: It seems like you changed the intent of this test, and somehow changed the behavior.
The intent here was to ensure that people cannot delete a field that is a `displayField`.  
That probably should still be the case after this.

Issue #241, opened 2019-12-27
TypeScript Typings are off on transformEntriesToType


bsgreenb:
pretty sure there’s a bug in the typescript typings for migration.transformEntriesToType https://github.com/contentful/contentful-migration/blob/master/index.d.ts#L267

The docs say that it should have sourceContentType and targetContentType fields https://github.com/contentful/contentful-migration#transformentriestotypeconfig

Think this prevents writing typescript migrations

![image](https://user-images.githubusercontent.com/980217/71534060-f9e34e00-28b0-11ea-8046-c331ec38b90b.png)

ruderngespra:
Hey, thanks again @bsgreenb, I added the fix.

PR #243, opened 2020-01-03
fix(types): Update typechecks for ITransformEntriesToTypeConfig

ruderngespra:
This is supposed to be a fix for #241.


bsgreenb:
don't you also need `identityKey` function field? https://github.com/contentful/contentful-migration#transformentriestotypeconfig

ruderngespra:
@bsgreenb You are right, thank you for pointing that out and also for reporting this issue! I will update the fix accordingly.

PR #245, opened 2020-02-17
fix(defaults): Make shouldPublish property consistently default to false

jroehl:
## Summary

This sets the `shouldPublish` prop for migrations consistently to false as a default. 

## Description

Introduced helper file to handle the default values and publish handling centrally. 

## Motivation and Context

The behaviour of an `undefined` - `shouldPublish` property was not consistent across the functions and not working as [documented](https://github.com/contentful/contentful-migration#transformentriesconfig).

## Todos

-   [x] Implemented fix

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -98,6 +98,8 @@ export interface IValidation {
 
 export type WidgetSettingsValue = number | boolean | string | undefined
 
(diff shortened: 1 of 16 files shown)


jroehl:
Won't fix ATM, as this is considered a breaking change.

PR #316, opened 2020-04-03
chore: add circleci config

TimBeyer:
## Summary
Move to CircleCI


TimBeyer:
@makinwab Since we didn't have audits enabled here before, I'd say let's first proceed with the initial setup and move and then add this later if we want.

PR #320, opened 2020-04-03
chore: remove code coverage checks

TimBeyer:
Sadly our setup is somewhat broken and constantly generates wrong coverage numbers.
Those result in red checks, causing auto merges to fail.
Remove the coverage - which was unreliable anyway - and unblock automatic merging.


TimBeyer:
The tests are also a tiny bit faster now since we can more easily parallelize the several tests

PR #350, opened 2020-04-27
validation should be asset-block instead of asset-inline

micksi:
## Summary
embedded-asset-inline doesn't exists and should be embedded-asset-block instead

## Motivation and Context
migration fails when using "embedded-asset-inline" in validation

-   [x] Implemented feature


micksi:
@phoebeschmidt would really like this to be merged as we currently are manually overwriting the error in our node_modules to be able to run our migrations.

PR #366, opened 2020-05-19
fix: add index.ts and make package.json refer to it in main field

lehnerm:
## Summary

Adds an src/index.ts file exporting the cli's runMigration
function. Fixes eslint/no-unresolved linter errors in 
dependent projects and makes using the cli as a library easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ npm install contentful-migration
 ### Usage as a library
 
 ```javascript
-const runMigration = require('contentful-migration/built/bin/cli').runMigration
+const { runMigration } = require('contentful-migration')
 const options = {
   filePath: '<migration-file-path>',
   spaceId: '<space-id>',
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -142,7 +142,7 @@ npm install contentful-migration
 ### Usage as a library
 
 ```javascript
-const runMigration = require('contentful-migration/built/bin/cli').runMigration
+const runMigration = require('contentful-migration')
```
suevalov, line 145: I would avoid default export in favor of 

```
const { runMigration } = require('contentful-migration');
```

so we can make non-breaking changes if we want export something else

lehnerm: :+1:

suevalov:
We need to update `index.d.ts` to reflect a new method being exported.

lehnerm:
When this PR is ready, what is the process to release it @phoebeschmidt? I'm not allowed to merge.

phoebeschmidt:
@lehnerm Ah okay, I will merge now and with properly formed commit messages, semantic release will release it to npm. 

I'll ask my manager about adding other teams so they have rights to merge

PR #404, opened 2020-06-22
refactor: use built-in didyoumean2 typings

foray1010:
## Summary

I am the auther of didyoumean2, we provide built-in ts typings since v2, so I removed the custom typings file

## Description

This PR removes https://github.com/contentful/contentful-migration/blob/e0187e242f249b6d2699444afb506cf305be7b9c/typings/didyoumean2/index.d.ts

## Motivation and Context

Do not need to maintain your own typings, to avoid out-sync when update this package

## Todos

-   [x] Implemented feature


marcolink:
@foray1010 Thank you for your contribution 🎉

PR #416, opened 2020-06-29
feat(tags): Add create tag support

ruderngespra:
Implement create tag operation for tags. 

`editTag` and `deleteTag` with their corresponding validations will follow in different prs.

## Open questions

The migration tool is designed around content types, which means that everything (actions, intents, validations etc.) is supposed to relate to a specific content type. Tags don't.

The current approach reuses and expands the existing classes and methods that are scoped to content types to also work with tags. But this might be confusing and difficult to expand later if we want to add other entities. How should we move forward with this in the future?

## Done
- Implement mvp `createTag` intent.
- Add fetcher and validation

## Todos for following prs
- add check for duplicate names
- add check for valid id
- change offline api constructor to accept options object instead of positional arguments
- rename `groupsWith` method


src/lib/action/tag-save.ts
```diff
@@ -0,0 +1,25 @@
...
+    super()
+    this.tagId = tagId
+  }
+
+  getEntityId (): string {
+    return this.tagId
+  }
+
+  getEntityType (): string {
+    return 'TAG'
```
TimBeyer, line 17: Could you take this and the other entity types too, and turn it into an `enum`?

src/lib/intent-validator/tag-update.ts
```diff
@@ -0,0 +1,20 @@
+import SchemaValidator from './schema-validator'
+import * as Joi from 'joi'
+
+class TagUpdateStepValidator extends SchemaValidator {
+  // TODO: Find a way to remove these protected properties as they are not used here.
```
TimBeyer, line 5: What do you mean? The base class uses those properties to render the validation errors.

https://github.com/contentful/contentful-migration/blob/master/src/lib/intent-validator/schema-validator.ts#L34-L35

src/lib/intent/composed-intent.ts
```diff
@@ -134,6 +137,24 @@ export default class ComposedIntent implements Intent {
     return true
   }
 
+  isTagIntent (): boolean {
+    // TODO Is this a viable option? How can we be sure that composed
+    // intents are not a mix of ct intents and tag intents?
```
TimBeyer, line 142: I think in theory you can only compose intents together that lead to being combined into the same request.  
The concern is still valid in that one could make a mistake in the grouping logic and have those combined together.  
But I think otherwise this should not be a problem.

src/lib/intent/base-intent.ts
```diff
@@ -155,6 +155,22 @@ export default abstract class Intent implements IntentInterface {
...
+
+  isTagCreate (): boolean {
+    return false
+  }
+
+  isTagUpdate (): boolean {
+    return false
+  }
+
+  isTagIntent (): boolean {
```
TimBeyer, line 170: We should really refactor things to get rid of this pattern in the future.  
An intent nowadays has more properties that say what it is not, than what it is :)

src/lib/intent/tag-create.ts
```diff
@@ -0,0 +1,44 @@
...
+  isTagCreate () {
+    return true
+  }
+
+  getTagId () {
+    return this.payload.tagId
+  }
+
+  groupsWith (): boolean {
+    return false
```
TimBeyer, line 20: I guess from seeing this `groupsWith` actually only says whether the current one should be grouped with the previous one. If that's the case, maybe we should rename it to `groupsWithPrevious` or something like that.  

Just an idea, maybe @zcei has some opinions on this and remembers our design choices? 😄

src/lib/offline-api/index.ts
```diff
@@ -137,9 +150,11 @@ class OfflineAPI {
   private requestBatches: RequestBatch[] = []
   private contentTypeValidators: ContentTypePayloadValidator[] = []
   private locales: string[] = []
+  private modifiedTags: Map<String, Tag> = null
 
-  constructor (contentTypes: Map<String, ContentType> = new Map(), entries: Entry[] = [], locales: string[], editorInterfacesByContentType: Map<String, EditorInterfaces> = new Map<String, EditorInterfaces>()) {
+  constructor (contentTypes: Map<String, ContentType> = new Map(), entries: Entry[] = [], locales: string[], editorInterfacesByContentType: Map<String, EditorInterfaces> = new Map<String, EditorInterfaces>(), tags: Map<String, Tag> = null) {
```
TimBeyer, line 155: Shouldn't we also initialize the `tags` map to an empty map instead of `null`?

Also I think it's time to refactor this to use an object in the constructor instead of separate positional arguments.

src/lib/offline-api/index.ts
```diff
@@ -485,6 +500,68 @@ class OfflineAPI {
...
+
+    // Mutate version bump
+    tag.version = tag.version + 1
+
+    this.modifiedTags.set(id, tag)
+
+    // TODO Do we need the savedTags?
+    // this.savedTags.set(id, tag.clone())
+
+    // TODO What is validated here?
```
TimBeyer, line 542: I think here we'd validate the final shape of the payload against the schema.

So you'd define a schema for what the tags payload looks like (I guess right now just `name` being a string between X and Y characters long)  and validate those here if I recall correctly.

test/integration/migration.spec.js
```diff
@@ -598,4 +599,15 @@ describe('the migration', function () {
       }
     ]);
   }));
+
+  it('creates a tag', co(function * () {
```
TimBeyer, line 603: Feel free to rewrite these things to `async/await` when you come across them.

test/unit/lib/fetcher.spec.ts
```diff
@@ -494,4 +494,55 @@ describe('Fetcher', function () {
...
+        name: 'bar',
+        description: 'A content type for a person'
+      })
+    }, noOp, {})
+
+    const request = sinon.stub()
+    request
+      .withArgs({
+        method: 'GET',
+        url: `/tags?limit=${Fetcher.perRequestLimit}&order=sys.createdAt&skip=0`
```
TimBeyer, line 510: Does the tags API actually support `order` right now without erroring?

src/lib/intent/base-intent.ts
TimBeyer: Just a note to self for future changes though, nothing to do right now :)

src/lib/entities/tag.ts
```diff
@@ -0,0 +1,54 @@
+// import { cloneDeep } from 'lodash'
+import APITag from '../interfaces/api-tag'
+
+class Tag {
+  private _id: string
```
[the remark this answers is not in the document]
ruderngespra: Yes. I am also confused about it and admittedly just used this pattern to align with the `ContentType` class.

src/lib/fetcher.ts
```diff
@@ -124,6 +126,16 @@ export default class Fetcher implements APIFetcher {
     })
   }
 
+  async getTagsForEnvironment (intentList: IntentList): Promise<APITag[]> {
+    // Don't fetch tags if migration does not use any:
+    if (!(intentList.getIntents().some((intent) => intent.isTagIntent()))) {
+      return []
+    }
+
+    const tags = await this.fetchAllPaginatedItems<APITag>('/tags')
```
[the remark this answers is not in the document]
ruderngespra: Makes sense! I added a comment.

src/lib/intent-list/index.ts
```diff
@@ -119,14 +121,27 @@ class IntentList {
               // TODO: maybe a better handling
             }
           }
+          if (entityType === 'TAG') {
+            try {
+              const tag = await api.getTag(entityId)
+              await action.applyTo(tag)
+            } catch {
+              // TODO: maybe a better handling
```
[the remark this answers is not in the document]
ruderngespra: This is also a structure (and a comment) I just took over from the way we handle editor interfaces, where the `try/catch` had apparently previously been added to fix a broken test. But I see now that this structure does not make any sense for tags and will remove it!

src/lib/intent-validator/tag-update.ts
ruderngespra: Ah, that was my confusion. I thought these were properties only relevant to content types and overlooked that they are about how to render the validation errors. I will remove the comment.

src/lib/intent/composed-intent.ts
ruderngespra: Do you mean the internal grouping logic that relies on the `groupsWith` and the `endsGroup` methods? So with regard to this worry we should not add an extra check somewhere, but just be cautious when we implement those methods?

src/lib/intent/base-intent.ts
ruderngespra: :) I agree. But to get rid of this we would need different base-intents for different entities I guess ...

src/lib/intent/tag-create.ts
ruderngespra: This makes sense in my opinion, I am just a bit reluctant to make it part of this pr as it will modify another 20 files. I would suggest that I do this renaming afterwards in a separate pr?

ruderngespra: But if you disagree, I am also happy to change it and bundle everything together!

TimBeyer: Nope works for me.

src/lib/offline-api/index.ts
ruderngespra: Yes, I changed it to a map initialization. I will refactor the constructor in a different pr!

test/unit/lib/fetcher.spec.ts
ruderngespra: My manual tests suggested that it does, although the parameter is being ignored.

src/lib/action/tag-save.ts
ruderngespra: I updated it to use an `enum`

src/lib/offline-api/index.ts
ruderngespra: I implemented this second round of payload validations for tags. I am not sure if this makes too much sense for now however. As far as I understand and suspect, this second round of payload validations was implemented to check for cases such as content types having too many fields at the end of a migration, something which you would not be able to check during the original create content type intent. In the case of tags however, the updates are far less complex, the only thing that needs to be checked is if there is a correct `name` property. But maybe it still makes sense to have this check here for the future, when editing tags becomes less trivial? Or am I missing something?

src/lib/intent-list/index.ts
```diff
@@ -107,26 +109,35 @@ class IntentList {
...
       // Auto insert publish and save
 
       if (intent.shouldSave()) {
-        const save = new ContentTypeSaveAction(intent.getContentTypeId())
+        let save: any
+        if (intent.isTagIntent()) {
+          save = new TagSaveAction(intent.getTagId())
+        } else {
+          save = new ContentTypeSaveAction(intent.getContentTypeId())
+        }
```
zcei, line 140: Another hint that we need to rethink design around intents [as mentioned by Tim](https://github.com/contentful/contentful-migration/pull/416/files#r451394607). Ideally we would have an `intentType` getter or alike, then we could say

```typescript
switch (intent.type) {
  [Intent.Tag]:
    return new TagSaveAction(intent.getTagId())
  [Intent.ContentType]:
  [Intent.EditorInterface]:
    return ContentTypeSaveAction(intent.getContentTypeId())
}
```

src/lib/intent/composed-intent.ts
zcei: Yes, I would say so as well. The grouping is only there to keep the amount of HTTP requests as small as possible. 

Also: when a group of intents contains at least one tag intent, it should fulfull the `isTagIntent`. If you have mixed intent types in there, the e.g. `isContentType` check would succeed as well. 

So it's up to the grouping logic to be "colocated" with / know about the HTTP adapter for the available endpoints and thus how to group things.

src/lib/interfaces/request.ts
```diff
@@ -4,6 +4,11 @@ interface HttpRequest {
   headers?: object
   data?: {
     fields?: object
+    // TODO: This is currently really confusing: We are reusing the
+    // httpRequest interface which is taylored for cts and therefore
+    // expects only a fields property. But for a tag creation it
+    // should instead expect a name property.
+    name?: string
```
zcei, line 11: IIs there anything we can do about this? Maybe a discriminated union so we can share the same properties. but know it's actual sub-type?

Is it always mutually exclusive to have either `fields` or `name` in data?

src/lib/migration-chunks/validation/tag.ts
```diff
@@ -0,0 +1,114 @@
...
+      context.toBeCreated.delete(tagId)
+      context.deleted.delete(tagId)
+    }
+
+    // TODO
+    // if (intent.isTagDelete()) {
+    //   context.deleted.add(tagId)
+    //   context.remote.delete(tagId)
+    //   context.created.delete(tagId)
+    // }
```
zcei, line 110: What's the issue here?

src/lib/interfaces/request.ts
ruderngespra: Yes, it is mutually exclusive.

I will have a look at what a discriminated union is and if we can apply it here :)

src/lib/migration-chunks/validation/tag.ts
ruderngespra: This is leftover code I want to implement in the next pr, I forgot to remove it for this one. Sorry about that!

src/lib/interfaces/request.ts
zcei: But can't a content type have a name as well? 🤔

ruderngespra: Good point. :thinking: I will try and understand why and how this worked before without the `name` property as part of the interface.

ruderngespra: Okay, it seems like I just did not understand the typescript validation properly. Typescript complained, because the `toApi` method of the tag entity returns a type that will never have a `field` property, so I could not assign it to an optional type `data` which when used requires one of its optional properties to be part of the interface. So there of course actually is in general a `name` property for content type requests, which makes all of this less confusing. `fields` is just another additional optional property. I would suggest then to just leave it like this, but also add the optional parameters `description` and `displayField`.

ruderngespra: I updated it accordingly.

src/lib/action/editorinterface-update.ts
```diff
@@ -24,9 +24,9 @@ class UpdateEditorInterfaceAction extends EntityAction {
     this._widgetNamespace = widgetNamespace
   }
 
-  getEntityType (): string {
+  getEntityType (): EntityType {
     // TODO: enums
```
TimBeyer, line 28: TODO is done 👍

PR #423, opened 2020-07-06
Adjust e2e tests so that they work with a test space that has tags enabled

ruderngespra:
- Adjust e2e and integration tests + provide updated nock templates

### DIFF
--- a/test/end-to-end/content-transform.spec.js
+++ b/test/end-to-end/content-transform.spec.js
@@ -86,7 +86,7 @@ describe('apply content transformation', function () {
       .expect(assert.plans.actions.apply())
       .end(async function () {
         const res = await getEntries(SOURCE_TEST_SPACE, environmentId, 'newsArticle');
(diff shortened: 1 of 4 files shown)


test/end-to-end/content-transform.spec.js
```diff
@@ -86,7 +86,7 @@ describe('apply content transformation', function () {
       .expect(assert.plans.actions.apply())
       .end(async function () {
         const res = await getEntries(SOURCE_TEST_SPACE, environmentId, 'newsArticle');
-        const entriesWithoutSys = res.items.map(i => _.omit(i, 'sys'));
+        const entriesWithoutSysAndMetadata = res.items.map(i => _.omit(i, 'sys')).map(i => _.omit(i, 'metadata'));
```
TimBeyer, line 89: I think you can just do `_.omit(i, ['sys', 'metadata'])`

ruderngespra: Thanks, I updated this.

ruderngespra:
I added adjustments also for the integration tests!

PR #431, opened 2020-07-09
Feat/add edit and delete tag support

ruderngespra:
## Done
- Implement `editTag`.
- Implement `deleteTag`.
- Add more validation for more complex `editTag` and `deleteTag` use cases [here](https://github.com/contentful/contentful-migration/blob/master/src/lib/migration-chunks/validation/tag.ts)

## Todos
- tag id (and content type id) validations are currently missing. Should they be added?

## Notes for test setup
- New integration tests are actually only being checked if env var `NOCK_RECORD` is set to 1!

### DIFF
--- a/src/lib/intent-validator/tag-update.ts
+++ b/src/lib/intent-validator/tag-update.ts
@@ -1,6 +1,10 @@
 import SchemaValidator from './schema-validator'
 import * as Joi from 'joi'
 
+// TODO: We check this here and as part of the offline-api
+// validation. Which is the right place?
+const MAX_NAME_LENGTH = 256
+
 class TagUpdateStepValidator extends SchemaValidator {
   protected article = 'a'
   protected displayName = 'tag'
@@ -11,7 +15,7 @@ class TagUpdateStepValidator extends SchemaValidator {
 
   get schema () {
     return {
-      name: Joi.string().required()
+      name: Joi.string().max(MAX_NAME_LENGTH).required()
     }
   }
 }
--- a/src/lib/migration-chunks/validation/tag.ts
+++ b/src/lib/migration-chunks/validation/tag.ts
@@ -32,7 +32,7 @@ class DuplicateCreate implements TagValidation {
   }
 }
 
-class AlreadyExistingCreates implements TagValidation {
+class AlreadyExistingIdCreates implements TagValidation {
   message = tagErrors.create.TAG_ALREADY_EXISTS
   validate (intent: Intent, context: ValidationContext) {
     if (!intent.isTagCreate()) {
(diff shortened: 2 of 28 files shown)


src/lib/intent-validator/tag-update.ts
```diff
@@ -1,6 +1,10 @@
 import SchemaValidator from './schema-validator'
 import * as Joi from 'joi'
 
+// TODO: We check this here and as part of the offline-api
+// validation. Which is the right place?
+const MAX_NAME_LENGTH = 256
```
TimBeyer, line 6: Seems good to catch this here as early as possible.

src/lib/migration-chunks/validation/tag.ts
```diff
@@ -32,7 +32,66 @@ class DuplicateCreate implements TagValidation {
   }
 }
 
-class AlreadyExistingCreates implements TagValidation {
+class EditBeforeCreates implements TagValidation {
```
TimBeyer, line 35: Feels like all of these have a pretty close match to what content types do.
Let's keep them separate for now but keep in mind the opportunity for extracting this behavior for the next entity that needs this kind of logic.

src/lib/offline-api/index.ts
```diff
@@ -146,6 +147,17 @@ const saveTagRequest = function (tag: Tag): Request {
...
 }
 
+const deleteTagRequest = function (tag: Tag): Request {
+  return {
+    method: 'DELETE',
+    url: `/tags/${tag.id}`,
+    headers: {
+      'X-Contentful-Version': tag.version
+    },
+    data: tag.toApiTag()
```
TimBeyer, line 157: DELETE requests send no additional data.

src/lib/migration-chunks/validation/tag.ts
ruderngespra: Yes, I agree. Ideally we would have an EntityValidation which could be made used of by Tags and ContentTypes and whatever else is coming.

src/lib/offline-api/index.ts
```diff
@@ -570,6 +583,26 @@ class OfflineAPI {
...
+    this.assertRecording()
+
+    const tag = await this.getTag(id)
+    // Store clone as a request
+    this.currentRequestsRecorded.push(deleteTagRequest(tag.clone()))
+
+    this.modifiedTags.delete(id)
+    this.savedTags.delete(id)
+
+    // TODO Do we need this in the case of tags? What does it validate
```
TimBeyer, line 596: I think we can skip it. 
Not sure if we even have a hook for `DeleteContentType`, but I can imagine we would have wanted to validate that there are no existing entries for that CT in case of a CT deletion. Otherwise the API will reject that deletion.
For tags we have nothing that needs to be validated AFAIK so I think it can be skipped.

src/lib/intent-validator/tag-update.ts
ruderngespra: I agree, but just wonder if we then need an offline-api schema validation at all for tags? I see that it makes sense for content types, because MAX_FIELDS have to be checked, but in the case of tags there really only is the name property. I think this is still my main uncertainty, how the separation of concerns between these two rounds of validations really work.

PR #441, opened 2020-07-16
Add support for adding tags to entries

ruderngespra:
## Done
- Basic working implementation of `setTagsForEntries` tested by integration tests.
- Add unit tests for setTagsForEntries.
- Add e2e test.
- Double check that metadata can never be part of an entry of a space that does not have tags enabled. Possibly add a test that ensures that a copy of an entry without metadata does not have metadata either.
- Update docs

## Nice to have (possibly for different pr)
- Smarter validation (don't allow duplicate tags, non-existing tags etc.)
- Better plan messages that mention tag ids
- Refactor confusing naming

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ For the given content type, updates the tags that are attached to its entries ac
 
 - **`contentType : string`** _(required)_ – Content type ID
 - **`from : array`** _(required)_ – Array of the source field IDs
-- **`setTagsForEntry : function (entryFields, entryTags, apiTags): object`** _(required)_ – Transformation function to be applied.
+- **`setTagsForEntry : function (entryFields, entryTags, apiTags): array`** _(required)_ – Transformation function to be applied.
     - `entryFields` is an object containing each of the `from` fields.
     - `entryTags` is an array containing link objects of all tags
 already attached to the entry.
@@ -421,7 +421,7 @@ const departmentMapping = {
 migration.setTagsForEntries({
   contentType: 'news-article',
   from: ['department'],
-  setTagsForEntry: (entryFields: any, entryTags: tagLink[], apiTags tagLink[]): tagLink[] => {
+  setTagsForEntry: (entryFields: any, entryTags: TagLink[], apiTags: TagLink[]): TagLink[] => {
     const departmentField = entryFields.department['en-US']
     const newTag = apiTags.find((tag) => tag.sys.id === departmentMapping[departmentField])
 
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ const departmentMapping = {
 migration.setTagsForEntries({
   contentType: 'news-article',
   from: ['department'],
-  setTagsForEntry: (entryFields: any, entryTags: TagLink[], apiTags: TagLink[]): TagLink[] => {
+  setTagsForEntry: (entryFields, entryTags, apiTags) => {
     const departmentField = entryFields.department['en-US']
     const newTag = apiTags.find((tag) => tag.sys.id === departmentMapping[departmentField])
 
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@
       - [`createTag(id[, opts])`](#createtagid-opts)
       - [`editTag(id[, opts])`](#edittagid-opts)
       - [`deleteTag(id)`](#deletetagid)
+      - [`setTagsForEntries(config)`](#settagsforentriesconfig)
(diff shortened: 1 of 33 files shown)


README.md
```diff
@@ -389,6 +392,48 @@ Uses the same options as [`createTag`](#createtagid-opts).
...
 
+#### `setTagsForEntries(config)`
+
+For the given content type, updates the tags that are attached to its entries according to the user-provided `setTagsForEntry` function. For each entry, the CLI will call this function once, passing in the `from` fields, link objects of all tags that already are attached to the entry and link objects of all tags available in the environment. The `setTagsForEntry` function is expected to return an array with link objects for all tags that are to be added to the entry. If it returns `undefined`, the entry will be left untouched.
+
+**`config : Object`** – Content transformation definition, with the following properties:
+
+- **`contentType : string`** _(required)_ – Content type ID
+- **`from : array`** _(required)_ – Array of the source field IDs
+- **`setTagsForEntry : function (entryFields, entryTags, apiTags): object`** _(required)_ – Transformation function to be applied.
```
TimBeyer, line 403: I think the return type is `array` actually.

README.md
```diff
@@ -389,6 +392,48 @@ Uses the same options as [`createTag`](#createtagid-opts).
...
+
+const departmentMapping = {
+  'san-francisco': 'department-sf',
+  'london': 'department-ldn'
+}
+
+migration.setTagsForEntries({
+  contentType: 'news-article',
+  from: ['department'],
+  setTagsForEntry: (entryFields: any, entryTags: tagLink[], apiTags tagLink[]): tagLink[] => {
```
TimBeyer, line 424: The type should be capitalized: `TagLink`.
Also you forgot ` :` after `apiTags`.

examples/31-set-tags-for-entries.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function (migration) {
+  migration.setTagsForEntries({
+    contentType: 'article',
+    from: ['title'],
+    setTagsForEntry: (fields, entryTags, apiTags) => {
+      const newTag = apiTags.find(tag => tag.sys.id === 'old');
```
TimBeyer, line 6: I know it's just an example, but I', a bit confused why `newTag` finds the tag with the ID `old` in the example 🙂

src/lib/interfaces/entry-set-tags.ts
```diff
@@ -0,0 +1,5 @@
+export default interface EntrySetTags {
+  contentType: string,
+  from: string[],
+  setTagsForEntry (inputFields: any, entryTags: object[], apiTags: object[]): any
```
TimBeyer, line 4: Can we give those tags a proper type instead of `object`?

src/lib/action/entry-set-tags.ts
```diff
@@ -0,0 +1,55 @@
...
+    // TODO: Refactor currently confusing naming!
+    const entries: Entry[] = await api.getEntriesForContentType(this.contentTypeId)
+    const tags: Map<String, Tag> = await api.getTagsForEnvironment()
+    const apiTags = [...tags.keys()].map((tagId) => {
+      return { sys: { id: tagId, type: 'Link', linkType: 'Tag' } }
+    })
+
+    for (const entry of entries) {
+      const entryFields = _.pick(entry.fields, this.fields)
+      const entryTags = entry.tags
```
TimBeyer, line 29: Reviewing this from the top, so not sure where this is defined, but should we maybe preserve the `metadata` nesting instead of having `Entry.tags`?

src/lib/entities/entry.ts
```diff
@@ -8,13 +8,15 @@ class Entry {
   private _version: number
   private _fields: object
   private _publishedVersion?: number
+  private _tags?: object[]
```
TimBeyer, line 11: Ah it was defined just below 🙂.
Anyway, same concerns. I think it might be fine to keep `tags` as a top level key since the implementation details are hidden from the users, but eventually in the future we might want to nest this under a `Metadata` object if we introduce more things there.
Still I'd say we should use explicit typings here instead of `object[]`.

examples/31-set-tags-for-entries.js
ruderngespra: fair point :)

src/lib/entities/entry.ts
ruderngespra: I agree! Using tags was just the most straightforward solution for now, and I will update the typings.

PR #446, opened 2020-07-21
feat(cli): configurable per request limit

marcolink:
## Summary
In some cases, the default request limit of 100 is not sufficient. 
This PR exposes a configurable limit parameter.

## Description

A new parameter `requestBatchSIze` can be set on cli or config file (contentfulrc.json) level.

### DIFF
--- a/test/unit/lib/fetcher.spec.ts
+++ b/test/unit/lib/fetcher.spec.ts
@@ -30,7 +30,7 @@ describe('Fetcher', function () {
     request
       .withArgs({
         method: 'GET',
-        url: `/entries?limit=${Fetcher.perRequestLimit}&order=sys.createdAt&sys.archivedAt[exists]=false&sys.contentType.sys.id[in]=newsArticle&skip=0`
+        url: `/entries?limit=100&order=sys.createdAt&sys.archivedAt[exists]=false&sys.contentType.sys.id[in]=newsArticle&skip=0`
       })
       .resolves({
         skip: 0,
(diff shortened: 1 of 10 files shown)


test/unit/lib/fetcher.spec.ts
```diff
@@ -30,7 +30,7 @@ describe('Fetcher', function () {
     request
       .withArgs({
         method: 'GET',
-        url: `/entries?limit=${Fetcher.perRequestLimit}&order=sys.createdAt&sys.archivedAt[exists]=false&sys.contentType.sys.id[in]=newsArticle&skip=0`
+        url: `/entries?limit=100&order=sys.createdAt&sys.archivedAt[exists]=false&sys.contentType.sys.id[in]=newsArticle&skip=0`
```
ruderngespra, line 33: I wonder if it makes sense to use a const variable for this instead of hard coding 100 in all these instances?

marcolink:
@TimBeyer how do you feel about `requestBatchSize`?

test/unit/lib/fetcher.spec.ts
marcolink: @ruderngespra 
First I thought the same, but then I thought the rest of the test is also not parameterized.

marcolink:
> 👍
> 
> One follow up question: Do we need to add this option also to the cli tool or does it appear there automatically once the cli tool makes use of the current migration package?

@ruderngespra Good question, let's check once this PR is merged, ok?

PR #460, opened 2020-08-03
feat(tags): validation

ruderngespra:
Add validation to not allow attaching non-existing tags to entries.

We don't yet check for duplicate tags (which would be done via a joi schema check), because of scope and a bug in the joi version we are using.

### DIFF
--- a/src/lib/offline-api/index.ts
+++ b/src/lib/offline-api/index.ts
@@ -2,6 +2,7 @@ import { omit, compact, get } from 'lodash'
 import FieldDeletionValidator from './validator/field-deletion'
 import { ContentTypePayloadValidator } from './validator/content-type'
 import { TagSchemaValidator } from './validator/tag'
+import { EntryValidator } from './validator/entry'
 import { ContentType, EditorInterfaces } from '../entities/content-type'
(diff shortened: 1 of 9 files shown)


src/lib/offline-api/index.ts
```diff
@@ -332,17 +340,17 @@ class OfflineAPI {
...
+    this.savedContentTypes.delete(id)
 
     // TODO: Is the DeleteContentType hook actually being implemented?
-    for (const validator of this.contentTypeValidators) {
-      if (validator.hooks.includes(ApiHook.DeleteContentType)) {
-        const errors = validator.validate(ct, this.savedContentTypes.get(id), this.publishedContentTypes.get(id))
-        this.currentValidationErrorsRecorded = this.currentValidationErrorsRecorded.concat(errors)
-      }
-    }
+    // for (const validator of this.contentTypeValidators) {
```
TimBeyer, line 348: So should we remove this or not?

ruderngespra: Ah, this was added again by mistake, I think it should be removed, because it is not implemented. I will do that.

PR #490, opened 2020-09-08
fix(publish): respect entry state CHANGED for preserve value

marcolink:
## Summary

In case of a published entry but with unpublished changes, `shouldPublish: 'preserve'` would always resolve in a published entry. With these changes we check for pending changes and only publish if there are none.

Resolves #229, #223

### DIFF
--- a/src/lib/action/entry-transform-to-type.ts
+++ b/src/lib/action/entry-transform-to-type.ts
@@ -1,4 +1,5 @@
 import EntryTransformToType from '../interfaces/entry-transform-to-type'
+import shouldPublishLocalChanges from '../utils/should-publish-local-changes'
 import { APIAction } from './action'
 import { OfflineAPI } from '../offline-api'
 import Entry from '../entities/entry'
@@ -91,7 +92,7 @@ class EntryTransformToTypeAction extends APIAction {
 
       }
       await api.saveEntry(targetEntry.id)
-      if (this.shouldPublish === true || (this.shouldPublish === 'preserve' && entry.isPublished)) {
+      if (shouldPublishLocalChanges(this.shouldPublish, entry)) {
         await api.publishEntry(targetEntry.id)
       }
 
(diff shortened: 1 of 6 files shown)


src/lib/utils/should-publish.ts
```diff
@@ -0,0 +1,14 @@
+import Entry from '../entities/entry'
+
+export default function shouldPublish (shouldPublish: boolean | 'preserve', entry: Entry) {
+  if (shouldPublish === true) {
+    return true
+  } else if (shouldPublish === false) {
+    return false
+  } else if (shouldPublish === 'preserve') {
+    const hasUnpublishedChanges = entry.version - entry.publishedVersion > 2
```
phoebeschmidt, line 9: should it be `>=`? https://www.contentful.com/developers/docs/tutorials/general/determine-entry-asset-state/

phoebeschmidt (changes requested):
just one sanity check q about >, >=, I'm confused

src/lib/utils/should-publish.ts
```diff
@@ -0,0 +1,17 @@
...
+export default function shouldPublish (shouldPublish: boolean | 'preserve', entry: Entry) {
+  if (shouldPublish === true) {
+    return true
+  } else if (shouldPublish === false) {
+    return false
+  } else if (shouldPublish === 'preserve') {
+    if (!entry.publishedVersion || !entry.version || isNaN(entry.version)) {
+      return false
+    }
+    const hasUnpublishedChanges = !!entry.publishedVersion && entry.version > entry.publishedVersion + 2
```
phoebeschmidt, line 12: let's just add some sort of note or documentation about this somewhere. then looks good to merge

src/lib/utils/should-publish.ts
marcolink: @phoebeschmidt thanks for pointing out!
I double checked, and in this case it's a bit different, at the moment we call this function to check, we already updated the version field. I added some explanation to the function and also renamed it to something not so general.

phoebeschmidt (approved):

PR #568, opened 2020-11-23
chore: add commitizen to ease semver commits

anho:
## Summary

Ease generation of correct commit msgs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1074,6 +1074,101 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@commitlint/execute-rule": {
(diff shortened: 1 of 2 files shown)


ruderngespra (changes requested):
This seems like a nice tool, but could you explain why we have to add updated nock files and api key secrets and space ids?  Wouldn't that be a problem for a public repository?

anho:
@ruderngespra ouch, glad someone caught that. That was me accidentally pushing playground stuff. It's gone now.

suevalov (approved):

ruderngespra (approved):
:+1: :)

PR #572, opened 2020-11-25
Use cma plain client

anho:
## Summary

Updating to new major version of CMA SDK and using the plain client. This causes a breaking change in case users consume the `createManagementClient` function.

@ruderngespra is there anything left in regards to documentation inside this repo?

## Todos

-   [x] Implemented feature

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,3 +82,7 @@ workflows:
             - test-unit
             - test-integration
             - test-e2e
+          filters:
+            branches:
+              only:
+                - master
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -1,4 +1,5 @@
 import * as path from 'path'
+import type { AxiosRequestConfig } from 'axios'
 
 import chalk from 'chalk'
 import * as inquirer from 'inquirer'
@@ -15,6 +16,8 @@ import writeErrorsToLog from './lib/write-errors-to-log'
 import { RequestBatch } from '../lib/offline-api/index'
 import { getConfig } from './lib/config'
 import ValidationError from '../lib/interfaces/errors'
+import { PlainClientAPI } from 'contentful-management'
+import { trim } from 'lodash'
 
 class ManyError extends Error {
   public errors: (Error | ValidationError)[]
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -82,3 +82,7 @@ workflows:
             - test-unit
             - test-integration
             - test-e2e
+          filters:
```
anho, line 85: Just making PRs on Circle a bit faster.

test/integration/setup.js
```diff
@@ -8,5 +8,6 @@ const requiredEnvVars = [
 const undefinedEnvVar = (key) => typeof process.env[key] === 'undefined';
 
 if (requiredEnvVars.some(undefinedEnvVar)) {
-  throw new Error('Please define all necessary "CONTENTFUL_INTEGRATION_*" environment variables');
+  console.error('Please define all necessary "CONTENTFUL_INTEGRATION_*" environment variables');
```
anho, line 11: Just throwing errors was not yielding the actual message to me, I'd attribute it to mocha :)

src/bin/cli.ts
```diff
@@ -15,6 +16,8 @@ import writeErrorsToLog from './lib/write-errors-to-log'
 import { RequestBatch } from '../lib/offline-api/index'
 import { getConfig } from './lib/config'
 import ValidationError from '../lib/interfaces/errors'
+import { PlainClientAPI } from 'contentful-management'
+import { trim } from 'lodash'
```
marcolink, line 20: afaik, this is the only way to ensure only relevant imports.
```suggestion
import trim from 'lodash/trim'
```

marcolink (approved):
LGTM as far as I can judge :)

suevalov (commented):
wow, that's the way easier than I expected. is it all that we need to change? basically, almost nothing were used from `contentful-management` and this tool has hardcoded urls for all CMA methods, right?

anho:
@suevalov yes, I was surprised as well.

suevalov (approved):

ruderngespra (approved):
lgtm :+1: :)

PR #577, opened 2020-12-01
feat(EntryEditors): Add support for migrating entry editors [EXT-2359]

kdamball:
## Summary

Adding support for migrating [multiple editors in editor_interface](https://www.contentful.com/developers/docs/concepts/editor-interfaces/#custom-editor)

## Description

Adds a new method `configureEntryEditors`

## Motivation and Context

Could have rewritten the previous method `configureEntryEditor` but this would break any previous scripts relying on its intended behaviour.

### DIFF
--- a/src/lib/entities/content-type.ts
+++ b/src/lib/entities/content-type.ts
@@ -94,12 +94,14 @@ class EditorInterfaces {
   private _controls: APIEditorInterfaceControl[]
   private _sidebar?: APIEditorInterfaceSidebar[]
   private _editor?: APIEditorIntefaceEditor
+  private _editors?: APIEditorIntefaceEditor[]
 
   constructor (apiEditorInterfaces: APIEditorInterfaces) {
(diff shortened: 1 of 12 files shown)


src/lib/entities/content-type.ts
```diff
@@ -264,6 +276,9 @@ class EditorInterfaces {
     if (this._editor) {
       result.editor = this._editor
     }
+    if (this._editors) {
```
Jwhiles, line 279: A request with both editor and editors is invalid. So can we ensure that we only send one? I would probably lean towards ignoring `editor` if `editors` exists - but we should probably also log a warning

Jwhiles (changes requested):

Jwhiles (approved):

marcolink (approved):
Can't say much because I have never seen this part of the project.

But I like the additional typescript a lot :)

Issue #329, opened 2020-04-13
Safe default for "shouldPublish" property


kuy:
## Expected Behavior

[transformEntries](https://github.com/contentful/contentful-migration#transformentriesconfig) DSL preserves entity statuses without specifying `shouldPublish` property explicitly.

## Actual Behavior

`transformEntries` DSL publishes all entities in default.

I executed a migration script without passing `shouldPublish` property on master environment (of course I tested on staging environment, but I didn't noticed) and lots of entries have published unexpectedly. It was nightmare for me.

## Possible Solution

`shouldPublish` property should be `preserve` as a safe default for beginners.

## Environment

-   **Node Version**: v13.11.0
-   **Package Manager Version**: 6.14.4
-   **Operating System**: macOS 10.14.6
-   **Package Version**: 1.3.14 (`contentful --version`)

PR #588, opened 2020-12-09
fix(transformEntries): shouldPublish will 'preserve' by default

Vinz93:
## Summary

Fixes: https://github.com/contentful/contentful-migration/issues/329

## Description

transformEntries DSL was publishing all entries by default, now it will be 'preserve' by default

### DIFF
--- a/test/unit/lib/actions/entry-transform.spec.ts
+++ b/test/unit/lib/actions/entry-transform.spec.ts
@@ -99,7 +99,7 @@ describe('Entry Action', function () {
         'hawaii': 'haukea'
       }
     })
-    expect(batches[0].requests[2].data.fields).to.eql({
+    expect(batches[0].requests[1].data.fields).to.eql({
       name: {
         'en-US': 'jim!',
         'hawaii': 'aloha'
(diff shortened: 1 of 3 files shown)


test/unit/lib/actions/entry-transform.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('Entry Action', function () {
         'hawaii': 'haukea'
       }
     })
-    expect(batches[0].requests[2].data.fields).to.eql({
+    expect(batches[0].requests[1].data.fields).to.eql({
```
Vinz93, line 102: In this test we were updating two entries, since we were publishing by default, there were 4 requests[ (1 update + 1 Publish) * 2], now we just have two update requests

that's why I updated the assertion 👍

TimBeyer:
This is technically a breaking change.  
Could you mark it as such in the commit?

matthew-contentful (commented):
This change looks great. I worry about what will happen to the cli when we publish this change. Will it auto-update to this new major version? If so, the cli itself should move up one major version too, but will it?

https://github.com/contentful/contentful-cli/blob/master/lib/cmds/space_cmds/migration.js#L93

I don't know enough about the auto-update to predict what will happen...

matthew-contentful (approved):
LGTM

PR #698, opened 2021-04-06
feat: add custom headers support

z0al:
## Summary

Added support for passing `headers` option down to `contentful-management`. I also added `-H/--header` option to the CLI.

A similar option/flag will be added to other related libraries as well (e.g. contentful/contentful-export#562).

## Motivation and Context

To enable users to pass custom HTTP headers. This is especially useful when trying out alpha features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ module.exports = function (migration, context) {
 | accessToken       |            | string  | The access token to use                                     | true     |
 | yes               | false      | boolean | Skips any confirmation before applying the migration,script | false    |
 | requestBatchSize  | 100        | number  | Limit for every single request                              | false    |
+| headers           |            | object  | Additional headers to attach to the requests                | false    |
 
 ### Chaining vs Object notation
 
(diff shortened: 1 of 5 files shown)


marcolink (changes requested):
I think it would be good to also add `headers` to list of known args for the built-in cli https://github.com/contentful/contentful-migration/blob/master/src/bin/usage-params.ts

Although `contentful-migration` is part of `contentful-cli`, it'd be great to have the same set of cli args for single command cli.

z0al:
> I think it would be good to also add `headers` to list of known args for the built-in cli https://github.com/contentful/contentful-migration/blob/master/src/bin/usage-params.ts
> 
> Although `contentful-migration` is part of `contentful-cli`, it'd be great to have the same set of cli args for single command cli.
> 
> This is also relevant for import/export

I was under the impression these were deprecated. I added support for the CLI option in this PR and will do the same in contentful-export/import

marcolink (approved):
> I was under the impression these were deprecated. I added support for the CLI option in this PR and will do the same in contentful-export/import

And you were right. The README only offers usage through `contentful-cli migrate`. But it's still used independently.

PR #701, opened 2021-04-08
feat: support relationship type validation

giotiskl:
## Summary

* Adds support for the `relationshipType` validation, which is used to enable composition relationships between content types in Compose.
* The new validation is conditionally supported based on the headers the user has passed along with the request. The alpha feature header `assembly-types` must be present for the validation to work.
* This PR is based on #698, which should be merged first.

## Todos

-   [x] Implemented feature


marcolink (approved):
Looks good to me! 
Just wondering if we have a chance for fewer explizit `any` types?

giotiskl:
> Looks good to me!
> Just wondering if we have a chance for fewer explizit `any` types?

yup, took care of that, replaced the custom `SimplifiedValidationError` type with `Joi.ValidationErrorItem` which is now available to us and got rid of the explicit any castings.

marcolink (approved):
:shipit:

PR #784, opened 2021-06-15
build: bump joi

TimBeyer:
Bump Joi to the most recent version.

The `transformation` validation now has a less specific error message because we could no longer get access to the required data in the new version. 

Almost everything else should be running as before.


ruderngespra (approved):
Thanks for doing this :) I remember struggling with joi bumps in the mast so many times ...

anho:
@TimBeyer we should bump `@types/joi` too as a follow-up.

PR #803, opened 2021-06-25
docs: document trackingFieldId setting

Silhoue:
## Summary

Add `trackingFieldId` setting description to the README.


anho (changes requested):
`index.d.ts` needs to be adjust as well.

anho (approved):

PR #820, opened 2021-07-08
Update README.md

georgemillard:
This PR updates the readme regarding the `changeFieldControl` function, to include missing details about two optional settings: `showCreateEntityAction ` and `showLinkEntityAction `

## Summary

Adds details of additional options to `changeFieldControl` function in the readme.

## Description

The following options are now documented:
showCreateEntityAction
showLinkEntityAction

## Motivation and Context

It took me quite a while to find what these options were called so I could use them as they are currently missing from the docs, even though they work when used in migration scripts.


marcolink (approved):

marcolink:
@georgemillard thanks for your contribution 🎉

PR #854, opened 2021-08-20
feat(tags): support for public tags

marcolink:
## Summary

Introduce Tag `visibility` for `createTag`

In your migration script, you will be able to get a tag's visibility by calling the getter `visibility`. 
If not explicitly set a tag's visibility will be `private`.

### DIFF
--- a/examples/32-create-public-tag.js
+++ b/examples/32-create-public-tag.js
@@ -1,0 +1,5 @@
+module.exports = function (migration) {
+  migration.createTag('publicsampletag', {
+    name: 'public-marketing'
+  }, 'public');
+};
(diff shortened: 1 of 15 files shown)


examples/32-create-public-tag.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function (migration) {
+  migration.createTag('publicsampletag', {
+    name: 'public-marketing'
+  }, 'public');
```
TimBeyer, line 4: Did you do it this way because the `visibility` is not changeable later so we don't want it as a separate action?

TimBeyer:
Could you add the new parameter to https://github.com/contentful/contentful-migration/blob/master/index.d.ts#L461 as well?  Sadly the public typings for the migration DSL need to still be maintained manually.

examples/32-create-public-tag.js
makinwab: Can't speak for Marco but that makes sense.

TimBeyer (approved):

PR #858, opened 2021-08-27
Terminate with an error when a migration is aborted

rouanw:
## Description

This Pull Request changes the behaviour when you select `N` (No) when the script prompts you whether you want to apply a migration.

Previous behaviour:
- Log warning
- Exit with 0 exit code

New behaviour:
- Log warning
- Exit with a non-zero exit code / throw error, depending on how `terminate` has been configured.

I'll leave it to your discretion, but this is likely a breaking change and would require releasing a new major semver version.

## Motivation and Context

I'm plugging `contentful-migration` into our CLI-based migration tooling (using [umzug](https://github.com/sequelize/umzug)), so it's nice if aborting the migration prevents subsequent commands from running.  The current behaviour is particularly problematic when using an external migration tool, because it will interpret the run as a success and will therefore consider the migration as applied, even though it hasn't been.

A workaround would be to force the script not to prompt, using the `yes` option. I quite like the nice feedback the `contentful-migration` library gives you before applying a change, though, so it would be a shame to lose that. (Especially when testing migrations locally against a test environment.)

### Before:

131104157-4554efea-0fb0-4536-b77e-d861ebe51d4a

### After:

131104809-2769fffb-79a8-4184-acbd-9ca1dbb21ba3

### DIFF
--- a/src/bin/cli.ts
+++ b/src/bin/cli.ts
@@ -225,6 +225,7 @@ const createRun = ({ shouldThrow }) => async function run (argv) {
     }
   } else {
     console.warn(chalk`⚠️  {bold.yellow Migration aborted}`)
+    terminate(new Error('Migration aborted'))
   }
 }
 


rouanw:
@marcolink @TimBeyer what do you think of this change?

draconteus:
Is there any suggestion here? It would be great to handle migration abort somehow. With throwing an error or just resolving a simple `{ aborted: true }` response.

PR #875, opened 2021-09-03
chore(types): add type guard

ruderngespra:
Add typeguard to fix typescript error which is triggered on test runs:

```

src/bin/usage-params.ts:11:50 - error TS2345: Argument of type 'string | number' is not assignable to parameter of type 'string'.
  Type 'number' is not assignable to type 'string'.

11     const filePath = path.resolve(process.cwd(), args._[0])
                                                    ~~~~~~~~~
```

### DIFF
--- a/src/bin/usage-params.ts
+++ b/src/bin/usage-params.ts
@@ -8,7 +8,7 @@ export default yargs
   .usage('Parses and runs a migration script on a Contentful space.\n\nUsage: contentful-migration [args] <path-to-script-file>\n\nScript: path to a migration script.')
   .demandCommand(1, 'Please provide the file containing the migration script.')
   .check((args) => {
-    const filePath = path.resolve(process.cwd(), args._[0])
+    const filePath = path.resolve(process.cwd(), args._[0].toString())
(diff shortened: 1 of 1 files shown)


src/bin/usage-params.ts
```diff
@@ -8,12 +8,16 @@ export default yargs
   .usage('Parses and runs a migration script on a Contentful space.\n\nUsage: contentful-migration [args] <path-to-script-file>\n\nScript: path to a migration script.')
   .demandCommand(1, 'Please provide the file containing the migration script.')
   .check((args) => {
-    const filePath = path.resolve(process.cwd(), args._[0])
-    if (fs.existsSync(filePath)) {
-      args.filePath = filePath
-      return true
+    if (typeof args._[0] === 'string') {
+      const filePath = path.resolve(process.cwd(), args._[0])
```
marcolink, line 12: wouldn't `args._[0].toString()` be sufficient?

ruderngespra: 👍🏻 :)

marcolink (approved):

PR #881, opened 2021-09-10
chore(docs): highlight that rtf references are not supported.

ruderngespra:


malin-sofrone (approved):
Looks good. Thank you Janko.

PR #883, opened 2021-09-10
docs: Add "app" to changeFieldControl's widget namespaces

ilyasmez:
## Summary

Add the `'app'` option to the possible widget namespaces of `changeFieldControl`.

## Description

As you can see [here](https://github.com/contentful/contentful-migration/blob/e7a2f95c0cb225e92ecfb19a50e7dcf8de8f172e/index.d.ts#L186), `changeFieldControl` supports `'app'` as a possible value of the `widgetNamespace` param.
So this PR documents that.


marcolink:
@ilyasmez thanks for your contribution 🎉

PR #930, opened 2021-10-19
feat(types): changeFieldControl builtin overload

AndrewLeedham:
## Summary

Adds overloads for changeFieldControl so `widgetId` can be auto-completed when using the `'builtin'` `widgetNamespace`.

## Description

Adds 2 overloads one for `'builtin'` and the other for `'extension' | 'app'` which scope the `widgetId` field to either be one of the built in Contentful widgets from https://www.contentful.com/developers/docs/extensibility/app-framework/editor-interfaces/ or any string.

## Motivation and Context

Makes writing migration scripts quicker without needing to lookup the docs for the IDs of builtin widgets. Essentially just a DX improvement.

image


dpbeaumont:
This PR would be extremely useful when using Typescript migrations.

PR #961, opened 2021-12-06
feat(config): Add standard header

andreascful:
Add standard header to requests

Fix various linting issues


src/bin/lib/config.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path'
 import * as os from 'os'
+import { v4 as uuidv4 } from 'uuid'
```
marcolink, line 3: `uuid` only exists as a transitive dependency, please make it a direct dependency.

marcolink (approved):

PR #963, opened 2021-12-07
chore(linkContentType): add test to document current link validation

ruderngespra:
Add test coverage to document current behavior on how validation for linking Content Types works.

Current behavior: Content Types with link fields which reference non-existing entities can still be created via the migration tool. Ideally, this should be fixed in the future to only allow linking to existing content types.


marcolink (approved):
Thanks for validating this 🥇

PR #965, opened 2021-12-08
feat: add editor layout creation

veu:
## Summary

This PR implements the first part of field grouping migration focusing on creating and updating field groups in the editor layout.

## Description

This PR implements:
* Creating an editor layout in the editor interface
* Creating and updating field groups in the editor layout
* Updating field group controls in the editor interface
* Partially: deleting field groups (the actual deletion logic would have bloated this PR further)

**A note on the code:** This PR breaks an assumption of the current logic so I had to update it. Currently updates to the editor interface do not require information from the content type (that isn’t already supplied by the migration) so it’s not necessary to fetch the content type for editor interface updates. When creating an editor layout we need to know all field IDs currently in the content type so in this one case we do need the content type. To handle this I added the `requiresContentType` method for intents so editor interface updates can make the content type available.

## Motivation and Context

It wasn’t possible to reasonable fit the whole field grouping implementation into one PR so I am merging this one into a branch that already contains the documentation and will merge the branch into master after the following open topics have been addressed as well.

* Deletion of editor layout and field groups
* Field group ID changes
* Moving fields and field groups
* Validation of editor layout structure
* E2E tests


src/lib/migration-steps/index.ts
```diff
@@ -50,6 +51,92 @@ class Field extends DispatchProxy {
...
+
+  constructor (contentTypeId, instanceId, fieldGroupId, dispatch, props = {}, { dispatchUpdate }) {
+    super({ dispatchUpdate })
+
+    this.contentTypeId = contentTypeId
+    this.instanceId = instanceId
+    this.fieldGroupId = fieldGroupId
+    this.dispatch = dispatch
+
+    // Initialize from third argument
```
anho, line 121: Its the fifth argument :D

src/lib/intent/editor-layout/editor-layout-delete-field-group.ts
```diff
@@ -0,0 +1,45 @@
...
+    return false
+  }
+  shouldSave (): boolean {
+    return false
+  }
+  shouldPublish (): boolean {
+    return false
+  }
+  toActions () {
+    return [
```
anho, line 31: Why doesn't this save the editor interface?

anho (commented):
I think a thorough look from @TimBeyer will help.

src/lib/action/editor-layout/editor-layout-create-field-group.ts
```diff
@@ -0,0 +1,46 @@
+import { EditorInterfaces } from '../../entities/content-type'
+import { EntityAction, EntityType } from '../action'
+
+interface FieldGroupItem {
```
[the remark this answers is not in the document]
veu: Good catch. These are unused.

src/lib/intent/editor-layout/editor-layout-delete-field-group.ts
veu: I’ll leave it as is but in the next PR I’ll add a validation to disallow creating editor layout / group controls if they exist.

src/lib/entities/content-type.ts
```diff
@@ -257,6 +274,78 @@ class EditorInterfaces {
...
+
+  updateEditorLayoutFieldGroup (fieldGroupId: string, props: Partial<APIEditorLayoutFieldGroupItem>) {
+    const fieldGroup = findFieldGroup(this._editorLayout, fieldGroupId)
+
+    Object.assign(fieldGroup.item, pick(props, ['name']))
+
+  }
+
+  createGroupControls () {
+    this._groupControls = []
```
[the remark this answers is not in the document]
veu: I’ll leave it as is but in the next PR I’ll add a validation to disallow creating editor layout / group controls if they exist.

src/lib/intent/editor-layout/editor-layout-delete-field-group.ts
veu: Added

marcolink (approved):
Looks great!

TimBeyer (approved):
Looks really good so far

PR #972, opened 2021-12-17
Editor layout validation (first round) [DANTE-157]

veu:
## Summary

* Implement field group deletion (interface was already there)
* Add validation for existing / duplicate field group IDs
* Add validation for deleting unknown field group / deleting field group twice

## Todos

-   [x] Add example / integration test for field group deletion
-   [x] Add unit test for validations

### DIFF
--- a/examples/xx-delete-editor-layout-field-set.js
+++ b/examples/xx-delete-editor-layout-field-set.js
@@ -1,7 +1,7 @@
 // TODO: rename file before merging to master
 
 module.exports = function (migration) {
-  const Page = migration.editContentType('page');
-  const editorLayout = Page.editEditorLayout();
+  const page = migration.editContentType('page');
+  const editorLayout = page.editEditorLayout();
   editorLayout.deleteFieldGroup('seo');
 };
--- a/examples/xx-delete-editor-layout-field-set.js
+++ b/examples/xx-delete-editor-layout-field-set.js
@@ -1,0 +1,7 @@
+// TODO: rename file before merging to master
+
+module.exports = function (migration) {
+  const Page = migration.editContentType('page');
+  const editorLayout = Page.editEditorLayout();
+  editorLayout.deleteFieldGroup('seo');
+};
--- a/src/lib/entities/content-type.ts
+++ b/src/lib/entities/content-type.ts
@@ -1,4 +1,4 @@
-import { cloneDeep, find, filter, findIndex, pull, forEach, pick } from 'lodash'
+import { cloneDeep, find, filter, findIndex, pull, forEach, pick, update } from 'lodash'
 
 import {
   APIContentType,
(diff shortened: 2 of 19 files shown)


examples/xx-delete-editor-layout-field-set.js
```diff
@@ -0,0 +1,7 @@
+// TODO: rename file before merging to master
```
zcei, line 1: Could those rather be todos on the PR so we see that "x from y task are completed" hint?

examples/xx-delete-editor-layout-field-set.js
```diff
@@ -0,0 +1,7 @@
...
+
+module.exports = function (migration) {
+  const Page = migration.editContentType('page');
```
zcei, line 4: Nit: I think in other examples we're using lower-cased identifiers to indicate that they are instances rather than classes.

src/lib/entities/content-type.ts
```diff
@@ -304,9 +304,27 @@ class EditorInterfaces {
...
-  // @ts-ignore
   deleteEditorLayoutFieldGroup (fieldGroupId: string) {
-    // TODO: delete field group
+    const fieldGroup = findFieldGroup(this._editorLayout, fieldGroupId)
+    if (!fieldGroup) {
+      return
+    }
+
+    const parentPath = fieldGroup.path.slice(0, -1)
+    const groupIndex = fieldGroup.path[fieldGroup.path.length - 1] as number
```
zcei, line 314: Do we have a recent enough TypeScript version that we could use `fieldGroup.path.at(-1)`? ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at))

src/lib/migration-chunks/validation/editor-layout.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+interface EditorLayoutValidation {
+  validate (intent: Intent, context: ValidationContext): string | string[]
+}
+
+class DuplicateCreate implements EditorLayoutValidation {
+  validate (intent: Intent, context: ValidationContext) {
+    if (!intent.isFieldGroupCreate()) {
+      return
```
zcei, line 22: Shouldn't those guard clauses return empty arrays according to the interface above?

src/lib/migration-chunks/validation/editor-layout.ts
```diff
@@ -0,0 +1,163 @@
...
+  editorInterfaces.forEach((editorInterfaces, ctId) => {
+    const editorLayout = editorInterfaces.getEditorLayout()
+    if (editorLayout) {
+      remote = remote.concat(collectFieldGroupIds(editorLayout).map(id => `${ctId}.${id}`))
+    }
+  })
+  const toBeCreated = intents.filter((intent) => intent.isFieldGroupCreate()).map(getScopedFieldGroupId)
+
+  let context = {
+    remoteFieldGroups: new Set(remote), // all currently (in the current iteration step) existing field groups
```
zcei, line 113: I think those descriptions would be better kept at the interface definition above, so we get them as hints in editors whenever we hover over them.

Plus `let context: ValidationContext` maybe?

src/lib/migration-parser.ts
```diff
@@ -96,7 +96,7 @@ const createMigrationParser = function (makeRequest: Function, config: ClientCon
       throw new errors.EditorInterfacesFetchingError()
     }
 
-    const existingEditorInterfaces: Map<String, EditorInterfaces> = new Map()
+    const existingEditorInterfaces: Map<string, EditorInterfaces> = new Map()
```
zcei, line 99: ❤️

examples/xx-delete-editor-layout-field-set.js
IBratWurst: It makes sense to add that in the final pr targeting master. For now we need to note it somewhere else I think

src/lib/migration-chunks/validation/editor-layout.ts
IBratWurst: Idk why ts does not complain (maybe because our return type is a super set of the interface?) but `undefined` as possible return looks better for me. Anyway yes, an empty string or array would behave the same.

src/lib/entities/content-type.ts
IBratWurst: Yes, we do 👍🏼

zcei (approved):

PR #982, opened 2021-12-30
feat: replace listr with listr2

patrickedqvist:
## Summary

This PR replaces the `listr` package (which hasn't received an update since October 2019) with the more maintained [listr2](https://github.com/cenk1cenk2/listr2)

## Motivation and Context

Replacing this package hopefully resolves an issue that I've encountered myself when trying to contribute to the [contentful-userland/contentful-action](https://github.com/contentful-userland/contentful-action) where I've updated all the packages to more recent versions.

By updating the packages I've encountered the following problem.

```.bash
Error: Cannot find any-observable implementation nor global.Observable. 
You must install polyfill or call require("any-observable/register") with your preferred implementation, 
e.g. require("any-observable/register")('rxjs') on application load prior to any require("any-observable").
```

I've tried to resolve the issue in my fork by following the above recommendations but without any luck. Digging deeper into the issue and found that the problem lies with the listr package used by this project.


ruderngespra:
Hey @patrickedqvist, sorry this took so long, and thanks for the suggested change which makes a lot of sense to me!

PR #992, opened 2022-01-11
feat: validate editor layout [DANTE-407]

Silhoue:
Add a schema and a validator for the editor interface as a whole. Cover the first validation error (more than 2 levels of nesting).

### DIFF
--- a/src/lib/offline-api/validator/errors.ts
+++ b/src/lib/offline-api/validator/errors.ts
@@ -84,6 +84,11 @@ const errors = {
       }
     }
   },
+  editorLayout: {
+    FIELD_GROUP_LEVEL_TOO_DEEP: () => {
+      return 'You cannot create a field group deeper that 2 levels'
+    }
+  },
   entry: {
     REQUIRED_PROPERTY: (path) => {
       return `The property "${path}" is required on an entry.`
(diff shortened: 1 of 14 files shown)


src/lib/offline-api/validator/errors.ts
```diff
@@ -84,6 +84,11 @@ const errors = {
       }
     }
   },
+  editorLayout: {
+    FIELD_GROUP_LEVEL_TOO_DEEP: () => {
+      return 'You cannot create a field group deeper that 2 levels'
```
IBratWurst, line 89: nit: impersonal messages are used for all other errors but TAGS_DO_NOT_EXIST

src/lib/offline-api/validator/schema/schema-validation.ts
```diff
@@ -55,6 +56,27 @@ const validateContentType = function (contentType: ContentType): PayloadValidati
...
+
+  const { error } = validateResult
+
+  if (!error) {
+    return []
+  }
+
+  return error.details.map((err): PayloadValidationError => {
+    if (err.type === 'alternatives.match') {
+      return {
```
IBratWurst, line 72: Shouldn't it return an array according to the function type?

IBratWurst (approved):

src/lib/offline-api/validator/errors.ts
Silhoue: updated

PR #994, opened 2022-01-12
Fix README.md in the editField section

joaopcm:
## Description

The function description says that the `id` parameter will be used to identify the field to be deleted.

## Motivation and Context

This change will fix and improve the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,7 +572,7 @@ Creates a field with provided `id`.
 
 Edits the field of provided `id`.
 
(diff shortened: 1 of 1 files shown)


joaopcm:
@andreascful could you give a hand here to get some reviews, please?

andreascful (approved):
Sorry for the late response. I was sick the last days.
Thank you for the change. Looks good to me. :)

PR #997, opened 2022-01-13
feat: change field group id and delete editor layout

IBratWurst:
## Summary
Functions to change field group id and delete editor layout

### DIFF
--- a/src/lib/intent/editor-layout/editor-layout-change-field-group-id.ts
+++ b/src/lib/intent/editor-layout/editor-layout-change-field-group-id.ts
@@ -4,36 +4,36 @@ import chalk from 'chalk'
 import { EditorLayoutChangeFieldGroupIdAction } from '../../action/editor-layout/editor-layout-update-field-group'
 
 export default class EditorLayoutChangeFieldGroupIdIntent extends Intent {
-  isEditorInterfaceIntent() {
+  isEditorInterfaceIntent () {
     return true
   }
-  isEditorLayoutUpdate() {
+  isEditorLayoutUpdate () {
     return true
   }
-  isGroupable() {
+  isGroupable () {
     return true
   }
-  groupsWith(other: Intent): boolean {
+  groupsWith (other: Intent): boolean {
     return other.isGroupable()
       && other.isEditorLayoutUpdate()
       && this.isSameContentType(other)
   }
-  endsGroup(): boolean {
+  endsGroup (): boolean {
     return false
   }
-  shouldSave(): boolean {
+  shouldSave (): boolean {
     return true
   }
-  shouldPublish(): boolean {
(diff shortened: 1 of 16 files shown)


src/lib/intent/editor-layout/editor-layout-change-field-group-id.ts
```diff
@@ -0,0 +1,55 @@
...
+      && this.isSameContentType(other)
+  }
+  endsGroup(): boolean {
+    return false
+  }
+  shouldSave(): boolean {
+    return true
+  }
+  shouldPublish(): boolean {
+    return false
```
anho, line 28: Are updates to editor interfaces immediate?

src/lib/migration-steps/action-creators.ts
```diff
@@ -330,6 +343,23 @@ const actionCreators = {
...
+        callsite: {
+          file: callsite.getFileName(),
+          line: callsite.getLineNumber()
+        }
+      },
+      payload: {
+        contentTypeId,
+        fieldGroupId,
+        props: {
+          newFieldGroupId
```
anho, line 359: What's the reason to put this into a props object?

anho (commented):
Beautiful, just minor clarifications.

src/lib/migration-steps/action-creators.ts
IBratWurst: Just to not pollute the type `RawStepPayload`, similarly to how it's used in field renaming

src/lib/intent/editor-layout/editor-layout-change-field-group-id.ts
IBratWurst: What do you intend as immediate? As far as I understood, all intents changes within a group are applied together.

anho: Immediate = they don't need to be published. I'm not sure about the lifecycle of editor_interfaces so just want to clarify.

src/lib/migration-steps/action-creators.ts
anho: Renaming a field seems to be the outlier in this file and usually all the info is put into the payload (as long as the keys are fixed).

src/lib/intent/editor-layout/editor-layout-change-field-group-id.ts
IBratWurst: Yep, no publication of editor interfaces

src/lib/migration-steps/action-creators.ts
IBratWurst: updated!

anho (approved):

PR #999, opened 2022-01-14
feat: validate editor layout [DANTE-407]

Silhoue:
Cover more validation errors for the editor layout:
- too many tabs
- too many field sets
- too few field groups
- tab with a field set control
- field set with a tab control

Plus fix the readme and how editor interface intents are grouped

### DIFF
--- a/src/lib/offline-api/validator/field-groups-count.ts
+++ b/src/lib/offline-api/validator/field-groups-count.ts
@@ -1,0 +1,34 @@
+import { ApiHook } from '..'
+import errorMessages from './errors'
+import { EditorInterfacePayloadValidator } from './editor-interface';
+import { EditorInterfaces } from '../../entities/content-type';
+import { collectFieldGroupIds } from '../../utils/editor-layout';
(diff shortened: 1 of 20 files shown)


src/lib/offline-api/validator/field-groups-count.ts
```diff
@@ -0,0 +1,34 @@
...
+    const tabsCount = editorLayout.length
+
+    if (groupsCount < 2) {
+      errors.push({
+        type: 'InvalidPayload',
+        message: errorMessages.editorLayout.TOO_FEW_FIELD_GROUPS()
+      })
+    }
+
+    if (groupsCount - tabsCount > 15) {
```
IBratWurst, line 25: Could we have these limits as outer constants and use them also in the related error messages?

Silhoue: Saved them as constants. They're still hardcoded in the readme files, so I left them in the tests as well.

IBratWurst (approved):

PR #1009, opened 2022-01-25
feat: align output and add E2E tests for field grouping

veu:
## Summary

Update field grouping migration.

## Description

* Add E2E tests
* Unify and group field grouping plan messages
* Fix: fetch content type when moving fields for field ID validation
* Fix: allow deleting a newly created field group
* Fix: throw error when attempting to create an editor layout if it already exists

## Motivation and Context

We added the field grouping migration in multiple steps. Instead of creating overhead by iterating on the output several times this PR aligns the output and adds matching E2E tests after finishing the functionality.


anho (commented):
LGTM from my side, I additionally tagged dev-workflows.

anho (approved):

PR #1017, opened 2022-01-31
feat: add more field group ID / name validation

veu:
## Summary

Add more validations for field group IDs and names to field grouping migration.

## Description

* Validate field group ID format when creating field group or changing field group ID
* Validate field group name length when updating field group
* Check if field group exists when changing group controls
* Update field group IDs used for validation when changing field group ID

### DIFF
--- a/src/lib/migration-chunks/validation/editor-layout.ts
+++ b/src/lib/migration-chunks/validation/editor-layout.ts
@@ -5,7 +5,6 @@ import errors from './errors'
 import { collectFieldGroupIds } from '../../utils/editor-layout'
 import EditorLayoutMoveFieldIntent from '../../intent/editor-layout/editor-layout-move-field'
 import { FieldsContext } from './index'
-import { EditorLayoutChangeFieldGroupId } from '../../intent'
 const editorLayoutErrors = errors.editorLayout
 
 const RELATIVE_MOVEMENTS = ['beforeField', 'afterField', 'beforeFieldGroup', 'afterFieldGroup']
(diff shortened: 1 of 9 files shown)


src/lib/migration-chunks/validation/editor-layout.ts
```diff
@@ -192,23 +238,23 @@ class DuplicateDeletes implements EditorLayoutValidation {
...
+
+    const contentTypeId = intent.getContentTypeId()
+    const fieldGroupId = intent.getFieldGroupId()
+    const newFieldGroupId = intent.getNewFieldGroupId()
 
     if (!fieldGroupId) {
       return changeErrors.MISSING_FIELD_GROUP_ID()
     }
-    if (!newFieldGroupId) {
+    if (newFieldGroupId === undefined) {
```
veu, line 254: Made the check more specific so empty string will trigger the _InvalidFieldGroupId_ validation.

src/lib/migration-chunks/validation/editor-layout.ts
```diff
@@ -89,6 +88,53 @@ class AlreadyExistingFieldGroupCreates implements EditorLayoutValidation {
   }
 }
 
+class InvalidFieldGroupId implements EditorLayoutValidation {
+  validate (intent: Intent) {
+    if (!intent.isFieldGroupCreate() && !intent.isFieldGroupIdChange()) {
```
veu, line 93: I decided to not group the validations by intent this time to avoid duplicating the checks.

anho (approved):

PR #1028, opened 2022-02-15
feat(options): Expose retryLimit parameter [ZEND-1819]

andreascful:
Expose the `retryLimit` parameter.
This allows clients to set how often to retry a request (e.g. in case of 429) so that they can easier overcome traffic peaks that would otherwise lead to some failing requests in the migration.


Vinz93 (approved):
🎉

PR #1060, opened 2022-04-19
feat: content type metadata support

IBratWurst:
## Summary

Introduce metadata to the tool and support for annotations management

## Description

Enable the tool to tolerate existing annotations on content type payloads.
Add functionalities to edit annotations assignments for content types and fields


marcolink (approved):

src/lib/intent/interface-example.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable unicorn/filename-case */
```
anho, line 1: Is this file a leftover?

src/lib/interfaces/annotation.ts
```diff
@@ -0,0 +1,127 @@
...
+  data: Record<string, any>
+}
+
+export type AnnotationAssignment = {
+  annotationId: string
+  target: AssignmentTarget
+}
+
+/**
+ * This data structure does not include all the properties defined in CEP-0125
```
anho, line 20: I think we can remove the notion of a CEP from a public repo.

anho (changes requested):
Awesome. Just minor stuff left.

IBratWurst:
> Awesome. Just minor stuff left.

🙏🏼  Thanks for spotting those

src/lib/action/field-rename.ts
```diff
@@ -29,11 +29,15 @@ class FieldRenameAction extends EntityAction {
     const fields = ct.fields
     const field = fields.getField(this.fieldId)
 
+    const fieldAnnotations = ct.getFieldAnnotations(this.fieldId)
     fields.deleteField(this.fieldId)
     field.id = this.props.newId
     delete field.newId
 
     fields.setField(this.props.newId, field)
+    if (fieldAnnotations) {
```
veu, line 38: Can’t hurt to add a test for this as well. E.g. so it doesn’t get overlooked in a refactoring.

IBratWurst: Is this it https://github.com/contentful/contentful-migration/pull/1060/files#diff-72b02b4b4284f14f9bfdfe3de28700e0469eb4d53066c70719c59a89846f3040 ? :)

veu: Yes, that works. I had only seen the examples up to this point.

anho (approved):

Issue #1029, opened 2022-02-15
Impossible to get the entry id using the transformEntries function


PierreCavalet:
## Expected Behavior
Being able to use `sys.id` and other info in the function used by `transformEntries`. 

For example:

```js
const myIdsToChange = ['a', 'b', 'c']

migration.transformEntries({
  contentType: 'newsArticle',
  from: ['sys.id'],
  to: ['myField'],
  transformEntryForLocale: function (fromFields, currentLocale) {
      if (currentLocale !== 'en-US') {
        return;
      }

      if (myIdsToChange.includes(fromFields.sys.id)) {
        return { myField: true };
      }
  }
});
```

Now I am aware "fromFields" variable name would be misleading. Maybe we could change it ?

## Actual Behavior
Currently it is not possible, we can only specify `fields` info, but not `sys` info. 

## Possible Solution

There are multiple solution to the problem. We could add a "fromSys" variable, we could always inject the entry ID, or we could allow the "pick" system to pick in the `sys` info even if it is not a field. 

Personally I would go towards the last one as it is very simple to implement, since we have all the information and we would just need to change the object in which lodash picks the properties:

https://github.com/contentful/contentful-migration/blob/168c7f2c02aab2caa3ac595c2dd136e2ed7609a0/src/lib/action/entry-transform.ts

replacing
```js
const inputs = _.pick(entry.fields, this.fromFields)

```
by
```js
const inputs = _.pick({id: entry.id, ...entry.fields}, this.fromFields)
```

would for example allow us to pick `id` and to use it in the `fromFields`.

I am open to suggestion and happy to contribute with a PR.

AlanZambrano-at:
Me and my team would definitely benefit from this feature!

@PierreCavalet your suggestion sounds just fine!

PierreCavalet:
A merge request is opened

PierreCavalet:
any news on that one ? @ruderngespra

PierreCavalet:
it's been 3 months, is anyone here ?

jameyel:
+1. I need this for all of the transform functions.

Crizzooo:
+1 we need this too

Can't we just pass `entry.sys` as the third argument to `transformEntryForLocale`?

We dont even need to allow users to pick from sys, sys will always be useful

ruderngespra:
[...]

PR #1074, opened 2022-05-03
feat: add entry id in the transformEntries field

PierreCavalet:
## Summary

Add entry id in the `transformEntries` function.

This pull request adds the possibility to select the entry id in the `transformEntries` function, retrieving it in the `fromFields` parameter. The goal is to able to use the id to transform entries, for example in the context of using a CSV file with entries ID and the values that need to be set for each entry on a specific field.

## Motivation and Context

An issue on the subject is open.
https://github.com/contentful/contentful-migration/issues/1029

### DIFF
--- a/src/lib/action/entry-transform.ts
+++ b/src/lib/action/entry-transform.ts
@@ -28,7 +28,7 @@ class EntryTransformAction extends APIAction {
     const entries: Entry[] = await api.getEntriesForContentType(this.contentTypeId)
     const locales: string[] = await api.getLocalesForSpace()
     for (const entry of entries) {
-      const inputs = _.pick(entry.fields, this.fromFields)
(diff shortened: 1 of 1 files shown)


ruderngespra:
Hey, hey, thanks for the contribution, I think it makes a lot of sense. We discussed this internally, but did not have time to prioritize looking into adding id as parameter to `transformEntries`. I will let you know once this changes!

marcolink:
Hi @PierreCavalet,
Thanks for your contribution - I think this makes a lot of sense!
If you could add a simple unit test for this, I'm happy to merge it.

Please let me know if you have time for it, if not, I"m happy to recreate it with tests and bring it to production :)

marcolink:
closing in favour of #1212

PierreCavalet:
I did not have time to contribute on this in between :) thanks for taking it !

PR #1081, opened 2022-05-17
Export type declarations to support use as library

bgschiller:
## Summary

Add the `"declaration": true` flag to tsconfig.json and tsconfig-dev.json.

## Description

This flag instructs the typescript compiler to output .d.ts files alongside the javascript in `built`. These can be used by typescript projects that consume contentful-migration as a dependency.

## Motivation and Context

I'm working on updates to deluan/contentful-migrate right now, and am
finding that I need to recreate a lot of type information that already
exists in this library. In addition to being a lot of work, duplicating
that information makes my work fragile. It can easily become out of sync
when the types inside contentful-migration change.

If you add this flag, the type declarations (.d.ts files) will become
part of the npm package, letting me depend on them without rewriting
them by hand.

In #213, @chrislambe comments:
> It looks like the interface we want is defined [in the source]. Any
> reason we can't just use that instead of duplicating it?

The answer to his question is that the declarations are not made
available in the npm package. This PR will fix his problem as well as
mine.

### DIFF
--- a/tsconfig-dev.json
+++ b/tsconfig-dev.json
@@ -5,6 +5,7 @@
     "target": "es2017",
     "esModuleInterop":true,
     "module": "commonjs",
+    "declaration": true,
     "lib": [
       "es2015",
       "es2016",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "target": "es2017",
     "esModuleInterop":true,
     "module": "commonjs",
+    "declaration": true,
     "lib": [
       "es2015",
       "es2016",


marcolink (commented):
@bgschiller thanks for your contribution 🎉 

could you also add a `types` field to the package.json?

bgschiller:
@marcolink I spent a little bit of time trying to figure out how to do that. As things stand, the hand-written index.d.ts at the root of the project repo is used because there's no "types" field in package.json. Without code changes, `built/index.d.ts` consists of only `export {}`, so it's not a good candidate to replace the hand-written one.

I still think this is worthwhile, but if we're replacing the hand-written index.d.ts we should try to do so without breaking changes. My PR to accomplish that is slightly larger: https://github.com/bgschiller/contentful-migration/pull/1/files

This PR as is (without "types") accomplishes what I need. I've already prepared the larger one, but I imagine it might take a bit more work to review so I left it separate.

PR #1096, opened 2022-06-21
change instances of cmaToken to managementToken

Oatelaus:
## Summary

Changes instances of `cmaToken` to `managementToken`

## Description

There appears to be a change done to the base API which uses `managementToken` instead of cmaToken. This means the access token must always be provided.

## Motivation and Context

This solves https://github.com/contentful/contentful-migration/issues/1095

PR #1102, opened 2022-07-20
chore(config-tests): refactor config tests

ruderngespra:
As preparation work for https://github.com/contentful/contentful-migration/pull/1096 we need to adjust our tests so that we can correctly document and test when `cmaToken` and `managementToken` should be picked up from `.contentfulrc`.

This is only adding tests and documenting current behavior, but not the fix yet.


Oatelaus:
@ruderngespra `cmaToken` is "Content **Management Token**" meaning that it's just another way of referring to "managementToken", no?

Cite: https://github.com/contentful/contentful-cli/pull/150#discussion_r297704401

marcolink:
Hey @ruderngespra, should we get it over the finish line?

ruderngespra:
> @ruderngespra `cmaToken` is "Content **Management Token**" meaning that it's just another way of referring to "managementToken", right?
> 
> Cite: [contentful/contentful-cli#150 (comment)](https://github.com/contentful/contentful-cli/pull/150#discussion_r297704401)

Yes, that's correct. Just an unfortunate renaming.

Vinz93 (approved):
LGTM 🙂

Issue #1111, opened 2022-08-29
Run migration API throws error on AWS Lambda environment


paweljankowski:
## Expected Behavior

Running `runMigration` function is possible in the AWS Lambda environment. 

## Actual Behavior

When running `runMigration` in the AWS Lambda environment we are getting `TypeError: Cannot read properties of undefined (reading 'getFileName')` error.  This is related to the [`callsites`](https://github.com/sindresorhus/callsites) library, which returns empty object for AWS Lambda env. 

## Possible Solution
Make optional chain of `callsite.getFileName()` -> `callsite?.getFileName()` and `callsite.getLineNumber()` -> `callsite?.getLineNumber()`.

## Steps to Reproduce

1. Create a simple AWS lambda function which will run Contentful migration.
2. Execute that function.
3. Check results.

## Context

We are using AWS Lambda as a backend solution. When user is creating new entity which represents Contentful's site, AWS Lambda function is creating content types which will be used for creating new entries by the API. 

## Environment
AWS Lambda with JS/Typescript

-   **Node Version**: 16.x 
-   **Operating System**:  AWS Lambda images

paweljankowski:
@Vinz93 Thanks for merging! :) Are you going to release new version on the npm repository?

PR #1112, opened 2022-08-29
Make optional chain for calling methods on callsite object

paweljankowski:
## Summary

Make an optional chain for calling methods on `callsite` object.

## Description
On the AWS Lambda execution environament `callsites` library just return empty object, so calling `getFileName()` or `getLineNumber()` is throwing errors.

PR addresses that issue by using the `?` operator before the method call. If the object is empty, then it won't fail. 

## Motivation and Context
Addresses [#1111](https://github.com/contentful/contentful-migration/issues/1111)

## Todos

-   [x] Implemented feature


Vinz93 (approved):
Hi @paweljankowski,

Thanks for reporting and solving this issue 🙂

PR #1113, opened 2022-09-22
Fix/do not write errors to file when used as library

simOneAudi:
## Summary

We are using this package to migrate our content models. Whenever a runtime error occurs in one of our backends based on Lambda functions, the whole handler errors out because of an unhandled promise rejection caused by the Lambda's filesystem being mounted as readonly and thus not allowing the error files to be written.

## Description

In order to fix this issue, the PR adjusts two things:
- Error files will not be written in case the function is used as a library/imported from another location (`shouldThrow === true`).
- The writing of error files will be awaited so that any possible exception raised will not result in an unhandled promise rejection.

## Motivation and Context

When used as a library, the error files are probably irrelevant, except for debugging purposes, or might - as in our case - even be harmful, as depending on the runtime environment, no filesystem is available.

## Todos

This is rather a proposal than an actual finished implementation, as I do not have the time and knowledge, to build and verify the changes.

-   [x] Provide proposal
-   [ ] Verify implementation
-   [ ] Add/adjust tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13046,9 +13046,9 @@
       }
     },
     "node_modules/shell-quote": {
-      "version": "1.7.2",
-      "resolved": "https://registry.npmjs.org/shell-quote/-/shell-quote-1.7.2.tgz",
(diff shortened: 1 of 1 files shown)


ruderngespra:
Hey, hey, sorry you are running into this, and thanks for the proposal! I think this makes a lot of sense.

I will discuss it internally and let you know once there is more to share regarding whether or not and when this will be implemented! Will close this PR for the time being.

PR #1124, opened 2022-10-04
feat: add resource link support and validation

veu:
## Summary

Add support for resource links.

## Description

* Add _ResourceLink_ field type
* Add `allowedResources` field property
* Validate resource link and allowed resource count
* Validate `allowedResources` structure

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -1,87 +1,73 @@
-# Contentful Migration – Validations
+Contentful Migration – Validations
+=====
 
 Here is a list of supported validations.
 
 ## INVALID_PROPERTY_NAME
-
 Whenever trying to set a property which doesn't exist, and we couldn't offer any suggestion to use as a replacement.
 
 **Example:**
-
 ```javascript
 field.fdskljhdf('Ctrl + arm')
 ```
 
 ## INVALID_PROPERTY_NAME_WITH_SUGGESTION
-
 Whenever trying to set a property which doesn't exist, and we could offer a suggestion to use as a replacement.
 
 **Example:**
-
 ```javascript
 field.nmae('First name')
 ```
 
 ## INVALID_PROPERTY_TYPE
-
 Whenever setting a value to a property which is not matching the expected type.
 
 **Example:**
-
 ```javascript
 field.name(true)
 ```
 
 ## INVALID_MOVEMENT_TYPE
-
 Whenever trying to move a field with the wrong type for the field id.
 
 Example:
-
 ```javascript
 contentType.moveField(3).toTheTop()
 ```
 
 ## INVALID_MOVEMENT_WITH_SELF
-
 Whenever trying to move a field relative to itself.
 
 Example:
-
 ```javascript
 contentType.moveField('name').afterField('name')
 ```
 
 ## INVALID_MOVEMENT_NAME
-
 Whenever trying to move a field relative to another with an invalid movement.
 
 Example:
-
 ```javascript
 contentType.moveField('name').xyz('surname')
 ```
 
 ## INVALID_MOVEMENT_NAME_WITH_SUGGESTION
-
 Whenever trying to move a field relative to another with an invalid movement but we can provide a suggestion.
 
 Example:
-
 ```javascript
 contentType.moveField('name').after('surname')
 ```
+
 
 While evaluating your migration script we validate that the actions you want to perform on content types and fields aren't wrong in any sense.
 These validations are bound to the entity type and action on which they ocurred. Therefore the error keys are scoped to it
 and have a layout of `{entity}.{action}.{validation name}`.
 
 ## field.create.FIELD_ALREADY_CREATED
-
 Whenever trying to create twice the same field.
 
 **Example:**
-
 ```javascript
 contentType.createField('name').name('First name')
 
@@ -89,22 +75,18 @@ contentType.createField('name').name('Last name')
 ```
 
 ## field.create.CONTENT_TYPE_DOES_NOT_EXIST
-
 Whenever trying to create a field on a non-existing content type.
 
 **Example:**
-
 ```javascript
 // While no `inexistent` content type exists in the target space
 migration.editContentType('inexistent').createField('foo')
 ```
 
 ## field.delete.FIELD_ALREADY_DELETED
-
 Whenever trying to delete a field twice.
 
 **Example:**
-
 ```javascript
 contentType.deleteField('name')
 
@@ -112,11 +94,9 @@ contentType.deleteField('name')
 ```
 
 ## field.delete.FIELD_DOES_NOT_EXIST
-
 Whenever trying to delete a field which doesn't exist.
 
 **Example:**
-
 ```javascript
 contentType.createField('name')
 // ...
@@ -124,22 +104,18 @@ contentType.deleteField('nmae')
 ```
 
 ## field.delete.CONTENT_TYPE_DOES_NOT_EXIST
-
 Whenever trying to delete a field of a content type which doesn't exist.
 
 **Example:**
-
 ```javascript
 // While no `inexistent` content type exists in the target space
 migration.editContentType('inexistent').deleteField('foo')
 ```
 
 ## field.update.FIELD_DOES_NOT_EXIST
-
 Whenever trying to edit a field which doesn't exist.
 
 **Example:**
-
 ```javascript
 contentType.createField('name')
 // ...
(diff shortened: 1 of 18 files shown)


src/lib/offline-api/validator/resource-links.ts
```diff
@@ -0,0 +1,54 @@
...
+import errorMessages from './errors'
+
+export default class ResourceLinksValidator implements ContentTypePayloadValidator {
+  public hooks = [ApiHook.SaveContentType]
+
+  public validate({ contentType }): (InvalidActionError | PayloadValidationError)[] {
+    const errors: (InvalidActionError | PayloadValidationError)[] = []
+
+    const fields = contentType.fields.toRaw()
+    const resourceLinkCount = fields.filter(isResourceLink).length
```
alicankargin, line 14: You can use the filtered array below then you would not need `isResourceLink(field) &&` in your if conditions.

```
const resourceLinks = fields.filter(isResourceLink)
if (resourceLinks.length > MAX_RESOURCE_LINKS) {
   ...
}

....

for (const resourceLink of resourceLinks) {
      if (!resourceLink.allowedResources) {
         ....
```

src/lib/offline-api/validator/schema/schema-validation.ts
```diff
@@ -292,6 +293,61 @@ const validateFields = function (
       }
     }
 
+    // Field `allowedResources` errors
+    if (path[1] === 'allowedResources') {
```
alicankargin, line 297: Is this the same as above? If so why not use the same condition for both?
```
if (prop.startsWith('allowedResources'))
```

alicankargin (approved):

Silhoue (approved):
Some nits, looks good otherwise

src/lib/offline-api/validator/resource-links.ts
veu: Then I’d need two filtered arrays and iterate over both which doesn’t seem like a better solution to what I’m doing now.

src/lib/offline-api/validator/schema/schema-validation.ts
veu: Either works. I can change it if you prefer this one.

docs/validation.md
```diff
@@ -225,97 +271,118 @@ const author = migration.createContentType('author').name('Author')
...
+  author
+    .createField('dummy-' + i)
     .name('Dummy field ' + i)
     .type('Symbol')
 }
 ```
 
+## contentType.TOO_MANY_RESOURCE_LINKS
+
+Whenever hitting the 3 resource link field limit on a content type, while trying to create a new field.
```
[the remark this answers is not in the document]
veu: This is gramatically correct.

src/lib/offline-api/validator/resource-links.ts
alicankargin: I don't understand why you would need 2 filtered arrays. It's just replacing the `fields` in the `for of` with `resourceLinks`.  It would be like the snippet below. Am I missing something?

```
    const fields = contentType.fields.toRaw()
    const resourceLinks = fields.filter(isResourceLink)

    if (resourceLinks.length > MAX_RESOURCE_LINKS) {
      errors.push({
        type: 'InvalidPayload',
        message: errorMessages.contentType.TOO_MANY_RESOURCE_LINKS(
          contentType.id,
          MAX_RESOURCE_LINKS
        )
      })
    }

    for (const field of resourceLinks) {
      if (!field.allowedResources) {
        errors.push({
          type: 'InvalidPayload',
          message: errorMessages.field.REQUIRED_DEPENDENT_PROPERTY(
            'allowedResources',
            field.id,
            'type',
            'ResourceLink'
          )
        })
      }

      if (field.allowedResources) {
        errors.push({
          type: 'InvalidPayload',
          message: errorMessages.field.FORBIDDEN_DEPENDENT_PROPERTY(
            'allowedResources',
            field.id,
            'type',
            'ResourceLink'
          )
        })
      }
    }

    ....
}

Comment (veu) [reply src/lib/offline-api/validator/resource-links.ts:14]: Your code creates a `FORBIDDEN_DEPENDENT_PROPERTY` error for resource links which doesn’t make sense. Check the original code: one `isResourceLink` has a `!` in front of it.

File: src/lib/offline-api/validator/schema/schema-validation.ts (hunk 2 above)
Comment (alicankargin) [reply src/lib/offline-api/validator/schema/schema-validation.ts]: I understand. I don't have a preference but I thought maybe the author of the snippet above did it for some reason. 

Let me explain my point. When you're reading the code from above you understand the first condition and then you come to the second condition which is cohesive to the first one but looks different. Then as a reader you have more to process. I think the reader would have less cognitive load if the conditions were written in the same style. 

If there is no reason for `const prop = fieldNames.join('.')` and `prop.startsWith('defaultValue')` personally I prefer `path[1]` because there is less to process.

File: src/lib/offline-api/validator/resource-links.ts (hunk 1 above)
Comment (alicankargin) [reply src/lib/offline-api/validator/resource-links.ts:14]: Yes you're right. I read it wrong.

File: src/lib/offline-api/validator/schema/schema-validation.ts (hunk 2 above)
Comment (veu) [reply src/lib/offline-api/validator/schema/schema-validation.ts]: Yes, I also prefer using `path[1]` because it’s more precise but I changed it to `startsWith` now to avoid touching the other code.

**Review by ruderngespra — APPROVED** [review]
This is awesome! 🎉 
I have a few questions / confusions, other than that very happy about this addition, I learned a lot about `resourceLinks` reading this PR.

In general, I understand correctly that this PR mainly "just" introduces one a new field type and its related additional validations which are encapsulated in the `ResourceLinksValidator`, correct?

File: docs/validation.md [hunk 4]
```diff
@@ -225,97 +271,118 @@ const author = migration.createContentType('author').name('Author')
...
 for (let i = 0; i < 60; i++) {
-  author.createField('dummy-' + i)
+  author
+    .createField('dummy-' + i)
     .name('Dummy field ' + i)
     .type('Symbol')
 }
 ```
 
+## contentType.TOO_MANY_RESOURCE_LINKS
```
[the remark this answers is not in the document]
veu: This is also required by the API so any content type fetched from the API will be below the limit.

docs/validation.md
```diff
@@ -434,130 +504,221 @@ author.createField('fullName')
...
   name: 'Settings'
-});
-
-editorLayout.editFieldGroup('settings')
-  .createFieldGroup('seo')
-  .name('SEO');
+})
 
-editorLayout.changeFieldGroupControl('seo', 'builtin', 'topLevelTab');
+editorLayout.editFieldGroup('settings').createFieldGroup('seo').name('SEO')
```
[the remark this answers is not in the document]
veu: I just realized that there’s quite a few unintended changes in this file. I’ll remove those.

examples/47-create-resource-link-fields.js
```diff
@@ -0,0 +1,30 @@
...
+    items: { type: 'ResourceLink' },
+    allowedResources: [
+      {
+        type: 'Contentful:Entry',
+        source: 'crn:contentful:::content:spaces/another-space',
+        contentTypes: ['contentType1', 'contentType2', 'contentType3']
+      }
+    ]
+  })
+}
```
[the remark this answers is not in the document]
veu: Correct. Just like for regular links we have two versions and I wanted to cover both.

src/lib/utils/resource-links.ts
```diff
@@ -0,0 +1,5 @@
+export const MAX_ALLOWED_RESOURCES = 3
+export const MAX_RESOURCE_LINKS = 3
+
+export const isResourceLink = (field) =>
+  field.type === 'ResourceLink' || (field.type === 'Array' && field.items?.type === 'ResourceLink')
```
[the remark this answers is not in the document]
veu: It’s weirdly named but it’s actually an object. Excerpt from example payload:
```json
"items": {
  "type": "ResourceLink",
  "validations": []
},
```

src/lib/intent-validator/field-update.ts
```diff
@@ -23,7 +23,8 @@ class FieldUpdateStepValidator extends SchemaValidator {
       omitted: Joi.boolean().required().strict(),
       deleted: Joi.boolean().required().strict(),
       linkType: Joi.string().required(),
-      items: Joi.object().required()
+      items: Joi.object().required(),
+      allowedResources: Joi.array().required()
```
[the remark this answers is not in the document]
veu: Yes, looks like this is wrong but why doesn’t this trigger any of the existing tests? Something should blow up, right?

docs/validation.md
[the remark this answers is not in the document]
veu: I can keep these changes or remove them but if I remove them we have to update the prettier rules because this rewrites the documentation every time: https://github.com/contentful/contentful-migration/blob/78e46f9f753749654c8361bc468f706b69a6d5fe/package.json#L155-L157

PR #1142, opened 2022-11-10
fix(editor-interface-controls): control changes of newly created fields should not trigger errors for editor interfaces with editorLayout property

ruderngespra:


andreascful (approved):
Very good concise change. Great that you figured out this hard issue!

mayakarabula (approved):

veu (approved):

src/lib/entities/content-type.ts
```diff
@@ -228,6 +228,25 @@ class EditorInterfaces {
       this._controls.push(control)
     }
 
+    // For existing editorInterfaces which use the editorLayout property, we need to check if the
+    // field is referenced, as the API requires every fieldId present in controls to also be
```
Silhoue, line 232:
```suggestion
    // field is referenced, as the API requires every fieldId present in the content type to also be
```
The `editorInterface.editorLayout` is updated on the backend whenever a field is added, deleted, or has the id changed. This solution may work as a quick fix, but long-term we should fetch the updated editor interface from the API instead of duplicating the logic here.

PR #1149, opened 2022-11-25
fix(types): change types to expose requestBatchSize []

andreascful:
## Summary

The `requestBatchSize` was not being exposed in all interfaces. Thereore users trying to use `runMigration' would not see it proposed as an option. This PR exposes the param and cleans up types a bit.

### DIFF
--- a/src/bin/lib/config.ts
+++ b/src/bin/lib/config.ts
@@ -1,21 +1,11 @@
 import * as path from 'path'
 import * as os from 'os'
 import { v4 as uuidv4 } from 'uuid'
+import type { ClientConfig } from '../../../index'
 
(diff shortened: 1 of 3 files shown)


src/bin/lib/config.ts
```diff
@@ -1,22 +1,12 @@
 import * as path from 'path'
 import * as os from 'os'
 import { v4 as uuidv4 } from 'uuid'
+import type { ClientConfig } from '../../../index'
```
marcolink, line 4: Since this exist before - I assume it has the right shape?

marcolink (approved):

mctrafik:
Thank you very much for this change.

PR #1156, opened 2022-12-13
Fix:  Updating Editor Layout when Content Type fields change 

Cyberxon:
## Summary

This PR is a follow up work for the previous [PR](https://github.com/contentful/contentful-migration/pull/1142) to make sure changes on the content type followed up by an editor interface/editor layout changes will not trigger any errors.

## Description

This PR adds an update to editor Layout when a field is created, updated and deleted in a content type with editor Layout.
This PR also solves an edge case of updating an editor interface followed with a delete action which triggered an error because the field was deleted from the editor layout while it should have still existed there, that's why we added the deepClone to send a copy of the editor layout at that point of time to the save API and avoid this case.

## Motivation and Context

This issue is a blocking issue to customers and it has risen to a severity 1 support ticket

### DIFF
--- a/examples/52-delete-field-in-content-type-with-editor-layout.js
+++ b/examples/52-delete-field-in-content-type-with-editor-layout.js
@@ -1,8 +1,8 @@
 module.exports = function (migration) {
-  const contentTypeWithEditorLayout1 = migration.editContentType('page')
+  const contentTypeWithEditorLayout = migration.editContentType('page')
 
-  contentTypeWithEditorLayout1.deleteField('anotherAdditionalField')
+  contentTypeWithEditorLayout.deleteField('anotherAdditionalField')
 
-  const editorLayout = contentTypeWithEditorLayout1.editEditorLayout()
+  const editorLayout = contentTypeWithEditorLayout.editEditorLayout()
   editorLayout.createFieldGroup('uploads').name('Uploads')
 }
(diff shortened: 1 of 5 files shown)


src/lib/entities/content-type.ts
```diff
@@ -467,6 +485,25 @@ class EditorInterfaces {
     }
   }
 
+  updateFieldIdInEditorLayout(oldFieldId: string, newFieldId: string) {
+    if (this._editorLayout?.length > 0) {
+      findEditorLayoutItem(
```
veu, line 490: Let’s not mix finding the item and renaming it. Like we do in the other methods we should first get the item and then change it.

examples/52-delete-field-in-content-type-with-editor-layout.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function (migration) {
+  const contentTypeWithEditorLayout1 = migration.editContentType('page')
```
Silhoue, line 2: Do we need these `1` and `2` in the names?

Cyberxon: no we don't, I just used that to debug why the migration wasn't running properly, thank you for noticing!

Silhoue (approved):

Issue #1095, opened 2022-06-21
Usage of cmaToken where managementToken is expected.


Oatelaus:
## Expected Behavior

When running the following command

`node_modules/.bin/ts-node node_modules/.bin/contentful-migration --space-id $CONTENTFUL_SPACE_ID --environment-id $CONTENTFUL_ENVIRONMENT $MIGRATION_FILE`

it would pick up the same `managementToken` which is displayed from `contentful config list` and do changes with that.

## Actual Behavior

When running the following command

`node_modules/.bin/ts-node node_modules/.bin/contentful-migration --space-id $CONTENTFUL_SPACE_ID --environment-id $CONTENTFUL_ENVIRONMENT $MIGRATION_FILE`

it throws an error about the access token not being defined.

## Possible Solution

There appears to be a change in `contentful-cli` which changed `cmaToken` to `managementToken` and due to this, the token is not being picked up.

## Context

This stop me from being able to implement migrations this in my repository and possibly CI.

## Environment

-   **Node Version**: v14.18.2
-   **Package Manager Version**: 6.14.15
-   **Operating System**: Darwin Kernel Version 21.3.0
-   **Package Version**: contentful-migration@4.9.1

ruderngespra:
Please let me know if this is still an issue after this https://github.com/contentful/contentful-migration/pull/1183

Oatelaus:
@ruderngespra looks good to me! I'm going to close this.

PR #1183, opened 2023-02-23
fix: use management token as fallback

ruderngespra:
Picks up the initiative from https://github.com/contentful/contentful-migration/pull/1096 and the related issue https://github.com/contentful/contentful-migration/issues/1095.

Use `managementToken` as fallback if no `cmaToken` is provided in `.contentfulrc`. We want to keep the `cmaToken` to ensure backwards compatibility, but allow for `managementToken` as well as it is the new default in `contentful-cli`.


marcolink (approved):
nice! 🎉

andreascful (approved):

PR #1190, opened 2023-04-17
feat(editor-interfaces): support update all editor interfaces in one step [SDK-2577]

Vinz93:
## Summary

Support `migration.updateAllEditorInterfaces()` migration step 

## Motivation and Context

We need a method that allows updating all the editor interfaces at the same time, the method will be required to include editor interfaces as part of the migration files generated by the merge app


Vinz93:
The method is not needed for merge app editor interfaces support

Issue #1170, opened 2023-01-27
transformEntries({ from: ['id'],... }) yields error rather than usable id


VanDoornP:
## Expected Behavior
I expect to be able to use the entry 'id' when I leverage`transformEntries`. For example:
```
  const migrationFunction = (migration) => {
    migration.transformEntries({
      contentType: 'myContentType',
      from: ['id'],
      to: ['fieldToTransform'],
      transformEntryForLocale(fromFields, currentLocale) {
        // Do stuff to update 'fieldToTransform'
      },
    });
  };
```

## Actual Behavior
I get an error in return:
```You cannot transform entries for content type "myContentType" because "id" can't be found on it.```

I have tried id, entryId, entry_id, contentfulId, contentful_id, _id, and sys.id. None of them worked.

## Possible Solution
Ensure the field 'id' can also be used, just like the other fields; or improve documentation if it's already possible.

## Context
I have a list of article entries I want to assign to pre-defined categories. `{ [articleId: string]: string }`. I can't do that unless I can obtain the id from the `fromFields` variable.

## Environment
-   **Node Version**: v18.12.1
-   **Package Manager Version**: v8.19.2
-   **Operating System**: #64~20.04.1-Ubuntu
-   **Package Version**: ?

VanDoornP:
Anyone?

PierreCavalet:
@VanDoornP I think it's a duplicate of my issue here: #1029 

I opened a PR. you won't be able to access it as of now, just so you know.

marcolink:
This has bee implemented: https://github.com/contentful/contentful-migration/releases/tag/v4.13.0

PR #1212, opened 2023-06-06
feat: provide entry id for transform and derive functions

marcolink:
## Summary

Provide the entry id for several transform callbacks.

## Description

For all transformation functions (`transformEntries`, `deriveLinkedEntries` and `transformEntriesToType`) we provide the entry id of the entry in scope in the call signature of the transform callbacks. This allows for cross checking the id with any other source, or more custom validations within scope.

We opted-in for an wrapping object as a third parameter to eventually enrich the callback with further data in the future.

**Example:** (more in the updated readme)
```js
  migration.transformEntries({
    contentType: 'newsArticle',
    from: ['author', 'authorCity'],
    to: ['byline'],
    transformEntryForLocale: function (fromFields, currentLocale, {id}) {
      
      if(id === '<specific-entry-id>') {
        return;
      }
      
      if (currentLocale === 'de-DE') {
        return;
      }
      const newByline = `${fromFields.author[currentLocale]} ${fromFields.authorCity[currentLocale]}`;
      return { byline: newByline };
    }
  });
};
```

## Motivation and Context

This is solves issues #1170, #1029,  and the super seeds [PR](https://github.com/contentful/contentful-migration/pull/1074). 

## Todos

-   [x] add tests
-   [x] Update readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,10 +285,10 @@ The transform function is expected to return an object with the desired target f
 - **`contentType : string`** _(required)_ – Content type ID
 - **`from : array`** _(required)_ – Array of the source field IDs
 - **`to : array`** _(required)_ – Array of the target field IDs
-- **`transformEntryForLocale : function (fields, locale): object`** _(required)_ – Transformation function to be applied.
+- **`transformEntryForLocale : function (fields, locale, {id}): object`** _(required)_ – Transformation function to be applied.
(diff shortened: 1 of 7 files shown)


ruderngespra (approved):
Very cool to see this tackled 🎉 

If I see this correctly, this is no breaking change as this id can be completely ignored in the callbacks and never has to be provided or passed through by customers directly?

Issue #1221, opened 2023-06-23
configureEntryEditors definition missing from index.d.ts


nschirmer:
The [README](https://github.com/contentful/contentful-migration/blob/v4.13.1/README.md) states that there is a `configureEntryEditors` method when editing a content type, however such definition does not exist within [index.d.ts](https://github.com/contentful/contentful-migration/blob/v4.13.1/index.d.ts)

PR #1222, opened 2023-06-23
feat(types): add configureEntryEditors definition

nschirmer:
## Summary

As mentioned in issue #1221, the type definition for `configureEntryEditors` is missing.

I also added `editor-builtin` as an option for `widgetNamespace` for these entry editors. I'm not entirely sure if `builtin` is still a possible choice for these, given that it does seem to require `editor-builtin`, but I've left the choice there.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,6 +207,12 @@ export interface IFieldGroupWidgetSettings {
   [setting: string]: WidgetSettingsValue
 }
 
+export interface IEntryEditor {
+  widgetNamespace: 'builtin' | 'extension' | 'app',
+  widgetId: string,
+  settings?: IEditorInterfaceOptions
+}
+
 export interface ContentType {
   id: string
   instanceId: string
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -199,6 +199,12 @@ export interface IFieldGroupWidgetSettings {
   [setting: string]: WidgetSettingsValue
 }
 
+export interface IEntryEditor {
+  widgetNamespace: 'editor-builtin' | 'builtin' | 'extension' | 'app',
+  widgetId: string,
+  settings?: IEditorInterfaceOptions
+}
+
```
bexux, line 207: nit: can we doc block these so we know what builtin, editor-builtin, etc mean? Not a blocker, just nice to have.

Raymond-Ly (approved):

jjolton-contentful (approved):
Thanks very much for the contribution @nschirmer

PR #1236, opened 2023-08-07
fix(docs): Fix `linkType` values & typos

DevlinRocha:
## Summary

* Fixed values for `linkType`
* Wrapped referenced types in backticks
* Fixed minor typo in documentation

## Description

In my experience with Contentful migration scripts, `linkType` does not share values with `Type` as the documentation states, but rather takes in a `string` value of either `"Asset"` or `"Entry"`.

I have wrapped referenced types in backticks to be consistent with how they are referenced earlier in the documentation.

## Motivation and Context

I am using Contentful and was reading through the documentation when I found these typos and changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -390,7 +390,7 @@ For the given (source) content type, transforms all its entries according to the
   - `locale` one of the locales in the space being transformed
   - `id` id of the current entry in scope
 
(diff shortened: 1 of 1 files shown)


marcolink (approved):
@DevlinRocha Thanks for your contribution - looks good to me 👍

PR #1251, opened 2023-09-25
feat(): support host option

andreascful:
Support `host` as a provided option (in addition to contentfulrc.json)


marcolink:
Does this also allow for `host` as a CLI parameter?

andreascful:
@marcolink 
Yes it does. However, this also requires some changes to contentful-cli which I will open the PR when the new version for contentful-migration is released

marcolink (approved):

PR #1255, opened 2023-10-11
feat: improve request throttling [ZEND-4197]

marcolink:
We currently don't have a good way of throttling big amount of requests. For other tools we already use pQueue or pThrottle to limit the number of requests/s. This is mostly relevant for migration with transform functions on entries, as the amount of affected entities can easily be > 1000. 

This changes  introduces a new CLI flag `request-limit` (defaults to `10`, as this is the smallest limit from our free offering) which describes the max amount of requests this tool initiates within a second.


ruderngespra (approved):
Nice.

We should expose this flag in `contentful-cli` then as well I guess.

marcolink:
> Nice.
> 
> We should expose this flag in `contentful-cli` then as well I guess.

Would like to first get a customer confirmation that it all works as expected

Issue #1267, opened 2023-11-20
Upgrade axios to remove all CVE vulnerabilities


kyrielia:
## Context

Axios has a number of CVEs associated with it:
https://security.snyk.io/package/npm/axios

Contentful-migration currently uses v1.5.1:
https://github.com/contentful/contentful-migration/blob/master/package.json#L55

Axios should be updated to the latest version as soon as possible to remove all vulnerability issues. Ideally this should be released as soon as possible.

jjolton-contentful:
hey @kyrielia thanks for reporting, this has since been resolved so I'll go ahead and close this issue now. as of #1452 and v4.26.3 contentful-migration is set up with the latest axios v1.7.9

PR #1268, opened 2023-11-20
Update axios to 1.6.2 to remove vulnerabilities

kyrielia:
## Summary

See: https://github.com/contentful/contentful-migration/issues/1267

## Description

Upgrade axios to latest available version

## Motivation and Context

Using contentful-migration introduces a vulnerable transitive dependency into any user that uses this library. This fix will mitigate this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@hapi/hoek": "^10.0.1",
-        "axios": "^1.5.1",
+        "axios": "^1.6.2",
         "bluebird": "^3.7.2",
         "callsites": "^3.1.0",
         "cardinal": "^2.1.1",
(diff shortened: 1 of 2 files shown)


kyrielia:
I've made a similar PR in contentful-management:
https://github.com/contentful/contentful-management.js/pull/2055

Not 100% sure, but I think this PR might have to wait until the fix in contentful-management is merged in? It looks like contentful-migration uses management:
https://github.com/contentful/contentful-migration/pull/1268/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L60

t-col (approved):

t-col:
thanks for your contribution @kyrielia ! you're right that the axios 1.5.1 dependency won't be entirely removed until we merge your other PR in contentful-management.js and then update the version of that dependency here. But we'll do that separately.

PR #1274, opened 2023-11-28
Update exported signature of MigrationFunction

royalrex:
## Summary

Changing of MigrationFunction type. The context parameter should not be declared as optional.

## Description

This PR removes the optional setting on the context parameter. This will hint typescript that the parameter will always be there. My research of your codebase indicates that the context will always be provided. 

I assume that your intent with marking the context as optional, is to not force other developers to have context parameter present in their migration function signature. But that is not what is happening. Typescript is okay with "missing" parameters, but marking it optional as is, will force developers to null check the context object before using. This seems to be not needed.

## Motivation and Context

Improves typescript type checking.

## Todos

-   [x] Implemented feature

Typescript before change with context parameter
image

Typescript after change with context parameter
image

Typescript after change with **no** context parameter
image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -665,4 +665,4 @@ export type MigrationContext = ClientConfig & {
 /**
  * The shape of the migration function that should be exported.
  */
-export type MigrationFunction = (migration: Migration, context?: MigrationContext) => void
+export type MigrationFunction = (migration: Migration, context: MigrationContext) => void


mgoudy91 (approved):
I took a look and came to the same conclusion and believe this is fairly safe (and correct).

Thanks for the contribution!

Issue #1332, opened 2024-07-11
When a migration fails, error is not correctly printed out. Displaying "no such file or directory" instead


AssisrMatheus:
## Expected Behavior
If there's a validation issue on a migration that is scheduled to run. The according validation errors should be shown, example:
![image](https://github.com/contentful/contentful-migration/assets/12276174/064a3cb2-0fd2-443f-8f5a-19fe74f74a2a)

## Actual Behavior
Instead, a misleading, incorrect error is showing up instead:
![image](https://github.com/contentful/contentful-migration/assets/12276174/118198de-bcc2-4d15-b1ba-d5352e78d179)

```sh
Error: ENOENT: no such file or directory, open 'file:///Users/assisrmatheus/dev/perimetre/optrust-app/contentful/migrations/2024-07-11-09-24_blockCarouselItemLabel.js'
    at Object.readFileSync (node:fs:453:20)
    at renderStepsErrors (/Users/assisrmatheus/dev/perimetre/optrust-app/node_modules/contentful-migration/src/bin/lib/steps-errors.ts:37:29)
    at run (/Users/assisrmatheus/dev/perimetre/optrust-app/node_modules/contentful-migration/src/bin/cli.ts:153:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at migrate (/Users/assisrmatheus/dev/perimetre/optrust-app/contentful/migrate.ts:97:7) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'file:///Users/assisrmatheus/dev/perimetre/optrust-app/contentful/migrations/2024-07-11-09-24_blockCarouselItemLabel.js'
}
```
I am not sure that this is mac only, but since it's the current OS I'm using. Might as well be the case. But can't discard it's also not happening on other OSs as well

This is how migrations are being ran:
```ts
const { runMigration } = await import('contentful-migration');
for (const migration of migrationsToRun) {
  const migrationModule = await import(migration.path);
  const migrationFunction = migrationModule.default;

  console.log(`Migration: ${migration.name}`);
  await runMigration({
    spaceId: process.env.CONTENTFUL_SPACE_ID,
    environmentId: process.env.CONTENTFUL_ENVIRONMENT,
    accessToken: process.env.CONTENTFUL_MANAGEMENT_API_ACCESS_TOKEN,
    yes: true,
    migrationFunction
  });
  console.log(`Migration ${migration.name} completed`);
}
```

[...]

PR #1333, opened 2024-07-11
fix: fixes `no such file or directory` when migration fails

AssisrMatheus:
## Description

Makes sure to check for a file url instead of a file path before providing it to `readFileSync`

## Motivation and Context

Fixes issue #1332


jjolton-contentful:
Thanks for the contribution @AssisrMatheus, will get this tested and released.

jjolton-contentful (approved):

PR #1458, opened 2025-01-10
feat: allow NER resource types in allowedResources

BasKiers:
## Summary

Allows custom `type` parameter in `allowedResources` to support NER

## Description

If the `type` starts with `Contentful:` we validate whether it is an allowed (internal) ResourceType and whether the additional properties are correct. If it is a custom `type` which is used by NER no additional properties are allowed.

### DIFF
--- a/test/unit/lib/offline-api/validation/payload-validation-resource-links.spec.ts
+++ b/test/unit/lib/offline-api/validation/payload-validation-resource-links.spec.ts
@@ -272,7 +272,7 @@ describe('payload validation (dependencies)', function () {
       ])
     })
 
-    it('should return an error when the type value is referencing a invalid type', async function () {
+    it('returns an error when the type value is referencing a invalid type', async function () {
       const existingCts = []
       const errors = await validateBatches(function (migration) {
         const lunch = migration.createContentType('lunch').name('Lunch').description('A Lunch')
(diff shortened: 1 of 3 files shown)


test/unit/lib/offline-api/validation/payload-validation-resource-links.spec.ts
```diff
@@ -194,35 +194,133 @@ describe('payload validation (dependencies)', function () {
...
             type: 'InvalidPayload',
             message:
               'Allowed resource at index 0 on the field "mainCourse" has an invalid property: "contentTypes" must contain at least 1 items.'
-          },
+          }
+        ]
+      ])
+    })
+
+    it('should return an error when the type value is referencing a invalid type', async function () {
```
Silhoue, line 275: nitpick: use "returns" or "should return" consistently

Silhoue (dismissed):

vikaskumr (dismissed):

test/unit/lib/offline-api/validation/payload-validation-resource-links.spec.ts
BasKiers: Yea, it's duplicated code vs my own code. I usually prefer `it('should return ..')` while the pre-existing code uses `it('returns an error')`.

I'll change it to match the wording already existent in the repo but don't think it's something we should strive to write consistently as long as the test name is clear

BasKiers: Looking at the repo there are [6 more tests](https://github.com/contentful/contentful-migration/blob/abc0a9d94aee4dba19ef4a6016ce0dbcb3ffcf1e/test/unit/lib/offline-api/validation/payload-validation-rich-text-resource-link-embeds.spec.ts#L148) that use the `'should return ..'` syntax, all written by me 🙈 I did not change them as it's not in scope of this PR.

BobHemphill76 (approved):

PR #1465, opened 2025-02-20
feat: add support for locale based publishing [TOL-2810]

chrishelgert:
## Summary

Support for using locale-based publishing is added to keep the status more granular.

https://github.com/user-attachments/assets/5b6c5e74-cd71-4c33-8248-0f69bf982346

### DIFF
--- a/src/lib/action/entry-transform.ts
+++ b/src/lib/action/entry-transform.ts
@@ -9,12 +9,13 @@ class EntryTransformAction extends APIAction {
   private fromFields: string[]
   private transformEntryForLocale: Function
   private shouldPublish: boolean | 'preserve'
-
+  private useLocaleBasedPublishing: boolean
   constructor(
     contentTypeId: string,
     fromFields: string[],
     transformation: Function,
-    shouldPublish: boolean | 'preserve' = 'preserve'
+    shouldPublish: boolean | 'preserve' = 'preserve',
+    useLocaleBasedPublishing: boolean = false
   ) {
     super()
     this.contentTypeId = contentTypeId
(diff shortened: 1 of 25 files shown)


marcolink (approved):
I don't fully understand the * key, but it generally looks good :)

src/lib/action/entry-transform.ts
```diff
@@ -57,8 +59,18 @@ class EntryTransformAction extends APIAction {
       }
       if (changesForThisEntry) {
         await api.saveEntry(entry.id)
-        if (shouldPublishLocalChanges(this.shouldPublish, entry)) {
-          await api.publishEntry(entry.id)
+        if (shouldPublishLocalChanges(this.shouldPublish, entry, this.useLocaleBasedPublishing)) {
+          if (this.useLocaleBasedPublishing) {
+            const localesToPublish =
+              this.shouldPublish === 'preserve'
```
jjolton-contentful, line 65: not at all sure about this, but was wondering if this would benefit from checking if `entry.fieldStatus` is present like is happening in other places

jjolton-contentful (approved):
Fantastic work on this, I know quite a few devs are going to be so excited to see this land!!

The publishing steps in both entry transformation and derivation are pretty similar, extracting the logic (or at least the locale extraction part) into a shared helper might reduce duplication and simplify maintenance, but that is just a nitpick and is absolutely not necessary for this changeset.

src/lib/action/entry-transform.ts
chrishelgert: it should be already checked that it exists at this point :)

2wce (approved):
LGTM

PR #1497, opened 2025-05-19
Create 56-create-first-field-group.js

VikingRock:
## Summary
Goal is to create a field group and add 2 fields to that group. 

## Description

* Starting with a new content type with 2 fields
* Content type has no existing field groups
* Goal is to create a field group and add 2 fields to that group

## Motivation and Context
Note that a new editor tab should be created for the changes to make effect. Otherwise, no changes will be applied.


elylucas (approved):
Approving to bring into our local repo and evaluate from there, thanks @VikingRock!

Issue #1480, opened 2025-05-10
add suport for metadata concepts


broman-pl:
add support for managing taxonomy/concept validations on content model definition

gilnimer:
Bump.

whitelisab:
Hi @broman-pl and @glinimer, thanks for this request. We've shared this with the team that owns this area. While we don't have a timeline on when it might be prioritized, we've made sure your feedback is on their radar.

gilnimer:
@whitelisab thanks for the quick response ☺️ 
Would you appreciate a PR if I put the time into it?

whitelisab:
@gilnimer I'm checking internally with the team that owns the taxonomy feature, we will keep you posted!

PR #1516, opened 2025-07-04
GROOT-1863: add taxonomy validations to content type

Deeds67:
## Summary

The goal of this PR is to introduce the functionality that allows users to create taxonomy validations on content types using the migration cli.

https://contentful.atlassian.net/browse/GROOT-1863

## Description

Adds three functions for the user to create/modify/delete taxonomy validations: 
* `addTaxonomyValidation`
* `clearTaxonomyValidations`
* `setTaxonomyValidations`

The `examples/56-set-taxonomy-validations.js` file shows how these functions can be called. I started implementing `setTaxonomyValidation` and then decided to also implement `addTaxonomyValidation` as I think it's more intuitive. Would appreciate feedback here on how usable these are and if I should change/remove some.

## Motivation and Context

Requested by customers: https://contentful.atlassian.net/jira/polaris/projects/CCS/ideas/view/3115661?selectedIssue=CCS-2711&issueViewSection=capture 

https://github.com/contentful/contentful-migration/issues/1480


package.json
```diff
@@ -83,7 +83,7 @@
     "@types/sinon-chai": "^3.2.4",
     "@types/yargs": "^17.0.28",
     "chai": "^4.2.0",
-    "chai-as-promised": "^8.0.0",
+    "chai-as-promised": "^7.1.1",
```
Deeds67, line 86: I downgraded this due to this issue: https://github.com/chaijs/chai-as-promised/issues/301#issuecomment-2160108977

src/bin/cli.ts
```diff
@@ -64,7 +64,10 @@ export const createMakeRequest = (
     strict: false
   })
 
-  const makeBaseUrl = (url: string) => {
+  const makeBaseUrl = (url: string, raw?: boolean) => {
+    if (raw) {
+      return `https://${host}${url}`
```
Deeds67, line 69: I added this `raw` boolean, because taxonomies live at `organization` level, and the `parts` array below always adds `space` & `environment`.

src/lib/fetcher.ts
```diff
@@ -151,6 +155,31 @@ export default class Fetcher implements APIFetcher {
     return tags
   }
 
+  async getOrganizationIdFromSpace(): Promise<string> {
```
Deeds67, line 158: For now this is only used in tests, but in the future we'll add the ability to create `Concepts` and `ConceptSchemes` and we'll probably need something similar for that, so I added it into the fetcher. Would be happy to move it to the test package though.

examples/56-set-taxonomy-validations.js
```diff
@@ -0,0 +1,31 @@
...
+  // Add taxonomy validations one by one
+  contentType
+    .addTaxonomyValidation('specific-concept', 'TaxonomyConcept', { required: true })
+    .addTaxonomyValidation('categories-scheme', 'TaxonomyConceptScheme')
+
+  // Clear all taxonomy validations
+  contentType.clearTaxonomyValidations()
+
+  // Replace all taxonomy validations with a new set
+  contentType.setTaxonomyValidation([
```
SleepingWhale, line 13: nit: would it make sense to make it plural `setTaxonomyValidations`, the same way as `clearTaxonomyValidations`?

src/bin/cli.ts
SleepingWhale: No strong opinion, just thinking out loud: what do you think about extracting this into two dedicated functions — makeOrgBaseUrl and makeSpaceBaseUrl?

SleepingWhale (commented):
Nice work 🙌
Just added a few comments, hope it helps.

src/bin/cli.ts
```diff
@@ -121,6 +130,7 @@ const createRun = ({ shouldThrow }) =>
     const makeRequest = createMakeRequest(client, {
       spaceId: clientConfig.spaceId,
       environmentId: clientConfig.environmentId,
+      organizationId: clientConfig.organizationId,
```
whitelisab, line 133: It looks like passing in an `organizationId` for a request is only utilized by our integration tests in order to create concepts and concept schemes, correct? I'm wondering if we could create those concepts and concept schemes in the testing environment without needing to change this logic that is used for all requests made by the tool. Or maybe it's ok to change this if we think we will extend the tool to support organization-level entities in the future?

whitelisab (commented):
Overall this looks really great; especially thank you for the comprehensive tests you've added! I've added a few comments for you to consider, please let me know if you have any questions or need clarification

src/lib/entities/content-type.ts
```diff
@@ -731,6 +733,65 @@ class ContentType {
...
+    } else {
+      // Add new validation
+      current.push(newValidation)
+    }
+
+    this.setTaxonomyValidations(current)
+  }
+
+  clearTaxonomyValidations() {
+    set(this, '_metadata.taxonomy', undefined)
```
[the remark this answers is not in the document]
Deeds67: Great catch, thank you very much! I've modified it to use `[]` instead and refactored the pruning logic to not prune the taxonomy metadata so that it is in fact cleared. Also added a test case to validate this

src/bin/cli.ts
Deeds67: It is only used by the tests for now, however it is definitely planned in the upcoming PRs to support org-level entities, since Taxonomy Concepts and Concept Schemes are at org level.

examples/56-set-taxonomy-validations.js
```diff
@@ -0,0 +1,31 @@
+module.exports = function (migration) {
+  const contentType = migration.createContentType('taxonomy-example').name('Taxonomy Example')
```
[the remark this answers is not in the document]
Deeds67: Great suggestion, I've added a comment at the top of the example file and changed the ids to make it clear that they need to exist first. Once I implement the creation of Concepts/Concept Schemes in the CLI, I'll refactor this example.

src/bin/cli.ts
whitelisab: Given that you're planning on implementing support for concepts and concept schemes in the future, we can keep it in this PR. Thanks for clarifying!

PR #1530, opened 2025-08-04
feat: add entryIds parameter to transformEntriesToType

patrykkurczyna:
## Summary

  Add optional `entryIds` parameter to `transformEntriesToType` method to enable filtering transformation to specific entries only.

  ### Description

  This PR introduces an optional `entryIds` parameter to the `transformEntriesToType` configuration object, allowing users to limit the transformation operation to a specific subset of entries
  rather than processing all entries of the source content type.

  ### Changes made:

  - Interface Update: Added `entryIds?: string[]` to the `TransformEntryToType` interface
  - Implementation: Updated `EntryTransformToTypeAction` class to accept and filter entries based on provided IDs
  - Testing: Added comprehensive test case to verify filtering functionality works correctly
  - Documentation: Updated `README.md` with parameter description and usage details

  ### Motivation and Context

  When working with large content types containing many entries, I needed to transform only a specific subset of entries rather than all entries. Previously, the
  `transformEntriesToType` method would process every entry of the source content type, which could be inefficient and potentially risky when only certain entries need transformation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,7 @@ For the given (source) content type, transforms all its entries according to the
 - **`sourceContentType : string`** _(required)_ – Content type ID of source entries
 - **`targetContentType : string`** _(required)_ – Targeted Content type ID
 - **`from : array`** _(optional)_ – Array of the source field IDs, returns complete list of fields if not configured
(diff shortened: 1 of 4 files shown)


elylucasctfl:
Hi @patrykkurczyna, just letting you know this is on our radar and we are discussing it on the team. Expect to hear back from us soon on getting this reviewed.

PR #1540, opened 2025-08-27
fix: add resource editor types to builtin editor type definition [DX-351]

chasepoirier:
## Summary

Add both `resourceLinkEditor` and `resourceLinksEditor` to the valid types for the built-in editor type definition.


elylucasctfl (approved):
🚀 , one suggestion would be to put "Closes #XXX" in the description of the PR so the GH issue is linked and closed when the PR merges automatically.

PR #1565, opened 2025-11-08
chore: update pipeline to GHA and publish to NPM [DX-506]

elylucasctfl:
## Todos

-   [x] Implemented feature


.github/workflows/check.yaml
ethan-ozelius-contentful: should we add `test:typescript-declaration` to this?  Kinda seems like a redundant command since it should be covered by `npm run build` 🤔  but if so, let's go ahead and remove it.

elylucasctfl: ya, its already done during build so no need to do it again, so i removed it

